PR #9, opened 2022-07-12
add basic deployment based on docker

Alex-Izquierdo:
Add a basic deployment with docker. 
Includes:
- environs library implementation for better management of environment variables
- customize some variables through environment variables. 
- update gitignore
- dockerfile and docker-compose file
- update "createuser" script to work in a non-interactive way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,7 @@
 *.db
 .env
 /venv
+
+# IDE's
+.idea
+.vscode
--- a/README.md
+++ b/README.md
@@ -37,4 +37,11 @@ Visit this url:
 
 You have set up the development environment.
 
+## Run the application with docker-compose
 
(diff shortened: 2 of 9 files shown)


benthomasson:
I get an error when running docker-compose up --build 

`app_1  | sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) unable to open database file
app_1  | (Background on this error at: https://sqlalche.me/e/14/e3q8)
docker_app_1 exited with code 1`

Alex-Izquierdo:
Hi @benthomasson the error is caused by docker (I've tested only podman). Docker-compose mounts a volume and changes the path for the sqlite DB in order to persist it. Docker has no ability to mount the volume with a non-root user on the fly, I've fixed it creating that folder before mounting. 

Now it should work. Could you try it again?

benthomasson (approved):

PR #24, opened 2022-07-20
point ansible-event requirement to the latest version of the main branch

Alex-Izquierdo:
At this stage of development I think it's better to point the ansible-events to the latest version of the main branch.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,9 +22,7 @@ install_requires =
     uvicorn
     websockets  # TODO(cutwater): Check if this is dependency is necessary
     environs
-
-    ansible-events
-
+    ansible-events @ git+https://github.com/benthomasson/ansible-events.git
 
 [options.extras_require]
 sqlite =


cutwater (approved):

cutwater (changes requested):
@Alex-Izquierdo Can you please describe use case of this?

benthomasson (commented):
I do this when I build containers, but I am not sure it should be package requirements.

cutwater:
@benthomasson I'm rather tentative on this change. It might be useful to simplify dev environment setup and since we don't have a formal release at the moment it's probably won't hurt, but in general I'd prefer to avoid specifying it in the dependencies list and make the decision per specific environment.

benthomasson:
I am not sure how it would work in dev environments.  I usually have both repos checked out and use  `pip install -e .` in the same venv.    I guess that would override the package requirements.

Alex-Izquierdo:
@cutwater 
Indeed, it is to facilitate development (and also testing). 
@benthomasson I think your workaround will work as long as you run the `pip install` in `ansible-events-ui` first. 

Anyway, I think this affects more to you than me, so if you prefer don't it, I close the ticket. 

Would you agree to enable it by default (at least temporally) in the docker-compose?

benthomasson:
Yeah use it in the docker-compose setup for now.

PR #28, opened 2022-07-25
Setup database migrations

cutwater:
* Create initial alembic configuration and directory structure.
* Configure alembic to use project settings and run auto formatters when
  migrations are generated.
* Add initial database migration.
* Add PostgreSQL service to the docker-compose environment.
* Update project to support only PostgreSQL database.
* Remove aiosqlite from package dependencies.
* Update documentation.
* Refactoring: Remove database initialization at module import.
* Refactoring: Remove dependency on synchronous driver.
* Refactoring: Replace `database_url` with `async_database_url`
* Refactoring: Move `database.py` and `models.py` modules into `db` package.

Issue: https://issues.redhat.com/browse/AAP-4847


tools/docker/docker-compose.yml
```diff
@@ -8,11 +8,20 @@ services:
       dockerfile: tools/docker/Dockerfile
     ports:
       - "8080:8080"
-    volumes:
-      - sqlite_db:/opt/app-root/src/sqlite
     environment:
       - AE_HOST=0.0.0.0
-      - AE_DATABASE_URL=sqlite:///./sqlite/test.db
```
Alex-Izquierdo, line 15: If sqlite will be remove, also you could update this line: 
https://github.com/benthomasson/ansible-events-ui/blob/main/tools/docker/Dockerfile#L19

tools/docker/docker-compose.yml
```diff
@@ -8,11 +8,20 @@ services:
...
-      - AE_ASYNC_DATABASE_URL=sqlite+aiosqlite:///./sqlite/test.db
+      - AE_DATABASE_URL=postgres+asyncpg://postgres:secret@postgres/ansible-events
+    depends_on:
+      - postgres
+
+  postgres:
+    image: "docker.io/library/postgres:14"
```
Alex-Izquierdo, line 18: Is there any special reason to use v14? this version is not available yet on RHEL AFAIK, if it's possible to use 13, that would make the things easier to our mates of AAP.

cutwater: Thanks, sure, we can use v13. There is no specific reason to use v14.

tools/docker/Dockerfile
```diff
@@ -16,10 +16,9 @@ RUN pip install ansible && \
     ansible-galaxy collection install benthomasson.eda
 COPY . $WORKDIR
 COPY --from=ui-builder /app/ui/dist/ ./ui/dist
-RUN mkdir sqlite && chown -R $USER_ID ./
```
Alex-Izquierdo, line 19: please, maintain the `chown -R $USER_ID ./ ` only the `mkdir` should be removed.

ansible_events_ui/config.py
```diff
@@ -9,9 +9,9 @@ class Settings(BaseSettings):
     host: str = "127.0.0.1"
     port: int = 8080
 
-    # FIXME(cutwater): Use only asynchronous sqlalchemy connection
-    database_url: str = "sqlite:///./test.db"
-    async_database_url: str = "sqlite+aiosqlite:///./test.db"
+    database_url: str = (
+        "postgresql+asyncpg://postgres:secret@localhost:5432/ansible-events"
```
benthomasson, line 13: Change the database name to `ansible_events`.

benthomasson (changes requested):

benthomasson (approved):

PR #36, opened 2022-08-02
fix postgres startup for docker compose

Alex-Izquierdo:
Add a healthcheck for postgres to avoid connection errors when the stack starts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ You have set up the development environment.
 
 ## Run the application with docker-compose
 
-Requires docker-compose installed. [See the documentation](https://docs.docker.com/compose/install/) for instructions.
+Requires docker-compose installed. [See the documentation](https://docs.docker.com/compose/install/) for instructions. (latest stable version is recommended)
 
 ```sh
 cd tools/docker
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo:
@cutwater

cutwater:
@Alex-Izquierdo AFAIK healthcheck has no effect on service startup in `docker-compose`. The `app` container itself must check explicitly for database readiness.

https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on

> There are several things to be aware of when using depends_on:

>    depends_on does not wait for db and redis to be “ready” before starting web - only until they have been started. If you need to wait for a service to be ready, see [Controlling startup order](https://docs.docker.com/compose/startup-order/) for more on this problem and strategies for solving it.

Alex-Izquierdo:
@cutwater I'm not sure, maybe the documentation is outdated. Or Maybe the [default condition ](https://github.com/compose-spec/compose-spec/blob/master/spec.md#long-syntax-1) is "service healthy". The last version from pip seems honors the healthcheck. You can check it. I did several tests and it works.

cutwater:
@Alex-Izquierdo I tested in with `docker-compose` v2.6.1 and it still fails, but looks like I was looking at the outdated specification indeed. With your changes and additionally updating `depends_on` as the following, it works as expected:

```
    depends_on:
      postgres:
        condition: service_healthy
```

Alex-Izquierdo:
@cutwater Ok. I added the condition and update the docs to recommend the latest version of docker-compose

cutwater (approved):

PR #37, opened 2022-08-02
add optional default user in migrations

Alex-Izquierdo:
Add an optional parameter in migrations to seed a default user and enables it in the docker-compose. 
Documentation updated

### DIFF
--- a/ansible_events_ui/db/migrations/script.py.mako
+++ b/ansible_events_ui/db/migrations/script.py.mako
@@ -8,6 +8,7 @@ Create Date: ${create_date}
 from alembic import op
 import sqlalchemy as sa
 ${imports if imports else ""}
+from alembic import context
 
 # revision identifiers, used by Alembic.
 revision = ${repr(up_revision)}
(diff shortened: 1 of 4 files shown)


Alex-Izquierdo:
@cutwater please review.

ansible_events_ui/db/migrations/script.py.mako
```diff
@@ -8,6 +8,7 @@ Create Date: ${create_date}
 from alembic import op
 import sqlalchemy as sa
 ${imports if imports else ""}
+from alembic import context
```
cutwater, line 11: I don't think that changes in the template are necessary. Normally database migrations will include only schema migrations, so in my opinion it is redundant to use this template for all generated revisions.

ansible_events_ui/db/migrations/script.py.mako
```diff
@@ -17,8 +18,15 @@ depends_on = ${repr(depends_on)}
 
 
 def upgrade() -> None:
-    ${upgrades if upgrades else "pass"}
-
+    schema_upgrades()
+    if context.get_x_argument(as_dictionary=True).get('create_default_user', None):
```
cutwater, line 22: `create_default_user` is a local option. It definitely shouldn't go into a template.

cutwater (changes requested):
@Alex-Izquierdo I don't think seeding database with default data for a development environment is a good idea. While I don't have strong arguments arguments against your approach, I have a feeling that it might get easily over- and wrongly used besides the development environment. 

* From the state of the database it is impossible track which options were use for applying each migration. 
* In general migrations should not heavily depend on an external parameters, as they tend to change, but applied migrations don't.
* Hardcoded weak password is a concern.

Alex-Izquierdo:
Ok @cutwater Let's discuss it with the team and I can close it or apply your changes.

Alex-Izquierdo:
Agreed to find a better way to create it.

PR #41, opened 2022-08-05
Fixes activation log/stdout

benthomasson:


ansible_events_ui/api.py
```diff
@@ -266,12 +266,12 @@ async def read_output(proc, activation_instance_id, db_session_factory):
     #   that is available in BackgroundTasks
     async with db_session_factory() as db:
         line_number = 0
-        # FIXME(cutwater): The `done` variable is never set to True.
         done = False
         while not done:
             line = await proc.stdout.readline()
             if len(line) == 0:
-                break
+                done = True
```
cutwater, line 273: @benthomasson 

It doesn't seem the boolean flag `done` is needed here. Why not simply:

``` 
while True:
    if not line:
        break
```

PR #42, opened 2022-08-05
Adds run_job

benthomasson:
Experimenting with ideas for AAP-5082


ansible_events_ui/api.py
```diff
@@ -507,6 +509,52 @@ async def list_job_instances(db: AsyncSession = Depends(get_db_session)):
     return result.all()
 
 
+@router.post("/api/job_instance/")
+async def create_job_instance(
+    j: JobInstance, db: AsyncSession = Depends(get_db_session)
```
cutwater, line 514: Please use more descriptive variables names. Single character variables should be avoided unless they're used as a loop index or their scope is narrow.

`j` is typically used as a loop index.

cutwater (changes requested):

ansible_events_ui/api.py
```diff
@@ -507,6 +509,52 @@ async def list_job_instances(db: AsyncSession = Depends(get_db_session)):
...
+    job_uuid = str(uuid.uuid4())
+
+    query = insert(job_instances).values(uuid=job_uuid)
+    result = await db.execute(query)
+    await db.commit()
+    (job_instance_id,) = result.inserted_primary_key
+
+    event_log = asyncio.Queue()
+
+    task = asyncio.create_task(
```
cutwater, line 535: I think we should document somewhere (if it hasn't been done already) that the tasking system implementation we use brings a limitation on number of worker processes that are used to run the application. The application is limited to only a single process mode and will have unexpected side effects if run in multi process mode.

cutwater:
@benthomasson 

I think we should document somewhere (if it hasn't been done already) that the tasking system implementation we use brings a limitation on number of worker processes that are used to run the application. The application is limited to only a single process mode and will have unexpected side effects if run in multi process mode.

benthomasson:
@cutwater  At this point, this server is only meant to work in single process mode.   In the future, multiple process mode can be supported if it is needed to meet some performance target that single process mode fails to support.

benthomasson:
@lgalis please review the UI portions of this PR.

lgalis (approved):

PR #44, opened 2022-08-08
Adding rules for writing fast db apps

benthomasson:
Adds my rules for writing fast database apps.

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -1,0 +1,43 @@
+
+
+Ben's rules for writing performant database applications
+
+These are lessons that I have learned over a decade of writing Django and other
+Python web applications dealing with large amounts of data.  They might be
+unusual and controversial, so that's why I am wrote them down here.  They are
+not rules that we must follow in this project.  They are a guide for if and
+when we need to scale to large amounts of data or have fast database
+operations.
+
+1. Think in sets and do bulk operations.  A single entry is a set of one.   The
+   cost of a single insert or update is similar to the cost of a bulk insert or
+   update since most of the cost of database writes are updating indicies.
+
+2. Know what SQL statement that you want to send to the database first.  Use
+   tools to generate that SQL second. ORMs generate really slow SQL unless you
+   are very careful and know what you are doing.
+
+3. Don't override the save function.  This prevents #1.  Don't use most of ORM
+   features since they conflict with #1.
+
+4. Get the data into the database as fast as possible without processing it.
+   Don't leave the user waiting for processing before responding that you
+   recevied the data.  Receive and write the data quickly and then process
+   it later.   Make separate tables for incoming data and processed data.
+   Don't add indicies to incoming data tables.   Provide and API or event
+   that lets them know that the data has been processed.
+
+5. Process the data in the database.  The database manipulates data many times
+   faster and scales better than Python applications.  Create and temporary
+   tables if needed and drop them when you are done with the operations.
+
+6. Create reporting tables. If you need to generate reports, create another set
+   of paritioned tables for those reports and ONLY do read, bulk inserts, and
+   drop table partition operations on them.  Don't report out of the relational
+   tables since it will be way too slow.
+
+7. Write correct queries first and optimize them second if needed.  Some queries
+   will be slow due to the structure of the relational data.  This can be fixed
+   by reorganizing the data into reporting tables.  Optimizing can take much longer
+   than writing the slow correct query and you'll need the slow query anyway to know
+   that your optimized query is correct.  Besides it is fun to get 10x and 100x speed-ups.
--- a/RULES.md
+++ b/RULES.md
@@ -14,11 +14,13 @@ operations.
    update since most of the cost of database writes are updating indicies.
 
 2. Know what SQL statement that you want to send to the database first.  Use
-   tools to generate that SQL second. ORMs generate really slow SQL unless you
-   are very careful and know what you are doing.
+   tools to generate that SQL second. Think about the number and the complexity
+   of queries you're writing. Avoid making queries in for loops.  Instead combine
+   those queries into one or two queries and process the data in a batch.
 
-3. Don't override the save function.  This prevents #1.  Don't use most of ORM
-   features since they conflict with #1.
+3. Create a diagram of the schema in Lucid charts, Miro, or Omnigraffle
+   something similar.  Database schemas that look organized in a diagram are
+   easier to understand when writing queries for them.
 
 4. Get the data into the database as fast as possible without processing it.
    Don't leave the user waiting for processing before responding that you
@@ -30,11 +32,14 @@ operations.
 5. Process the data in the database.  The database manipulates data many times
    faster and scales better than Python applications.  Create and temporary
    tables if needed and drop them when you are done with the operations.
+   That said, getting SQL to do the manipulation that you want can be quite
+   tricky.  Exporting batches of data to tools like Pandas can be useful
+   and scale horiztonally in background tasks.
 
-6. Create reporting tables. If you need to generate reports, create another set
-   of paritioned tables for those reports and ONLY do read, bulk inserts, and
-   drop table partition operations on them.  Don't report out of the relational
-   tables since it will be way too slow.
+6. Create reporting tables or databases. If you need to generate reports,
+   create another set of paritioned tables for those reports and ONLY do read,
+   bulk inserts, and drop table partition operations on them.  Don't report out of
+   the relational tables since it will be way too slow.
 
 7. Write correct queries first and optimize them second if needed.  Some queries
    will be slow due to the structure of the relational data.  This can be fixed


benthomasson:
@cutwater please review
@Ladas do you have anything that you'd like to add here?

RULES.md
```diff
@@ -0,0 +1,43 @@
...
+not rules that we must follow in this project.  They are a guide for if and
+when we need to scale to large amounts of data or have fast database
+operations.
+
+1. Think in sets and do bulk operations.  A single entry is a set of one.   The
+   cost of a single insert or update is similar to the cost of a bulk insert or
+   update since most of the cost of database writes are updating indicies.
+
+2. Know what SQL statement that you want to send to the database first.  Use
+   tools to generate that SQL second. ORMs generate really slow SQL unless you
```
cutwater, line 17: > ORMs generate really slow SQL

I disagree with this conclusion. People produce bad solutions unless they know exactly what they're doing. It's a matter of a competence, not a technology. 

However  Django and DRF experience may be harmful and makes the situation worse, due to Django's poor architecture. Django ORM gives a false impression of working with object model entirely and moves the focus away from the resulting SQL the ORM produces. DRF violates SRP and provides leaky abstractions (i.e. Serializers and Viewsets).

However we use the SQLAlchemy in the project, which is way closer to native SQL (unless used with implicit relationship loading techniques, that are not supported for asynchronous mode anyway). With SQLAlchemy 2.0 API the edge between Core and ORM becomes even thinner.
SQLAlchemy is simply a query builder, data mapper and a unit of work patterns, and I don't think there is anything wrong  in using them. In any more or less complicated project you'll end up building a data mapper on your own, unless taking an existing one.

I would put it differently:

Think about what you're doing. Think about number and complexity of queries you're writing. When deciding between processing data in Python application and the database, follow the advice nr. 5.

RULES.md
```diff
@@ -0,0 +1,43 @@
...
+   are very careful and know what you are doing.
+
+3. Don't override the save function.  This prevents #1.  Don't use most of ORM
```
cutwater, line 20: We don't have `save` function in SQLAlchemy. It's doesn't follow the Active Record pattern.

RULES.md
```diff
@@ -0,0 +1,43 @@
...
+   recevied the data.  Receive and write the data quickly and then process
+   it later.   Make separate tables for incoming data and processed data.
+   Don't add indicies to incoming data tables.   Provide and API or event
+   that lets them know that the data has been processed.
+
+5. Process the data in the database.  The database manipulates data many times
+   faster and scales better than Python applications.  Create and temporary
+   tables if needed and drop them when you are done with the operations.
+
+6. Create reporting tables. If you need to generate reports, create another set
```
cutwater, line 34: Or even standalone database or specialized database for analytical data processing.

cutwater:
@benthomasson Thank you for bringing this up. It's a nice idea to have guidelines especially when it comes to working with databases. I think some of your advises make perfect sense, however I found some of them somewhat controversial.

RULES.md
```diff
@@ -0,0 +1,48 @@
...
+   faster and scales better than Python applications.  Create and temporary
+   tables if needed and drop them when you are done with the operations.
+   That said, getting SQL to do the manipulation that you want can be quite
+   tricky.  Exporting batches of data to tools like Pandas can be useful
+   and scale horiztonally in background tasks.
+
+6. Create reporting tables or databases. If you need to generate reports,
+   create another set of paritioned tables for those reports and ONLY do read,
+   bulk inserts, and drop table partition operations on them.  Don't report out of
+   the relational tables since it will be way too slow.
```
Ladas, line 42: "relational tables": the better term here is probably "tables with raw data", since they are all tables in relational db I guess.

RULES.md
```diff
@@ -0,0 +1,48 @@
...
+2. Know what SQL statement that you want to send to the database first.  Use
+   tools to generate that SQL second. Think about the number and the complexity
+   of queries you're writing. Avoid making queries in for loops.  Instead combine
+   those queries into one or two queries and process the data in a batch.
+
+3. Create a diagram of the schema in Lucid charts, Miro, or Omnigraffle
+   something similar.  Database schemas that look organized in a diagram are
+   easier to understand when writing queries for them.
+
+4. Get the data into the database as fast as possible without processing it.
```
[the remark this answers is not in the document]
cutwater: I think it's also worth mentioning knowledge of a transaction scope and how the database operations work in concurrent environment.

RULES.md
cutwater: Or maybe normalized data. Raw data sounds a bit misleading.
The data in the database is not necessarily "raw", but optimized for either OLTP or OLAP processing systems.

RULES.md
```diff
@@ -0,0 +1,43 @@
...
+   features since they conflict with #1.
+
+4. Get the data into the database as fast as possible without processing it.
+   Don't leave the user waiting for processing before responding that you
+   recevied the data.  Receive and write the data quickly and then process
+   it later.   Make separate tables for incoming data and processed data.
+   Don't add indicies to incoming data tables.   Provide and API or event
+   that lets them know that the data has been processed.
+
+5. Process the data in the database.  The database manipulates data many times
```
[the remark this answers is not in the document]
Ladas: I think Pandas+Python is fine, you just need to make sure the batching is the right size, since the computation is all in memory. (working with large dataq sets). The other alternative for ETL would be data lake tools like S3+Trino that scales better and is way cheaper for large data sets. (Trino has SQL interface, so the rule still partially applies :-) )

cutwater:
I'd like to add a few items on database migrations.

@benthomasson @Ladas I'd appreciate your feedback.

## Writing migrations

#### Keep migrations isolated from the outside world

Migrations must not depend on environment they run in or configuration parameters used at the execution time. 

Migrations must not import any code that may change, including any functions, classes or constants from the project or any 3rd party projects.

When migrations are applied the database tracks only the identifier of the migration tree head or list of applied migrations. It's impossible to track the environment state used at the time each migration was applied and the environment may change. It leads to migrations failures or unexpected side effects that are hard to debug.

#### Use migrations for updating schema and existing data

Migrations can be also used for seeding the database, but only if the data introduced in the migrations doesn't depend on the environment and get updated rarely and via migrations. List of timezones is a good example of such constant data that may be stored in the database.

Be extremely careful with the data that is seeded via migrations, but may be updated by the business logic. In the subsequent migrations expect the data may have changed since the migration it was inserted in. Do not rely on auto-increment or randomly generated identifiers. Do not expect the data is present in the database.

For seeding the database with the data that relies on environment or configuration, use standalone scripts or tools. For example if you need to create an initial user records use scripts, management commands or tools like Ansible.

#### Keep migrations backwards compatible

In production environments where services upgraded without downtime or multiple versions of a service co-exist at the same time (i.e. A\B testing and Blue\Green deployments), applying a database migration must not break existing services.

Avoid changes that can break previous version of code (such as column and table renames and deletions) and provide an upgrade path if such change is required.

Ladas:
@benthomasson looks good overall, added some comments.

I'd add section for using DB constraints, to keep the DB integrity. The constraints provided by ORMs are not that good (e.g. ORM validation checks won't ensure the DB integrity 100%). Not sure how far Django/SQL Alchemy came with constraints, what is mapped to actual DB constraints and what not. And if foreign key constraint and cascade delete are using python implementation (slow) or are being translated to DB. (I just know that in Ruby world, the ORM got closer with the DB implementation after a decade of devs ignoring DB constraints entirely :-) )

@cutwater the migrations looks good I'd add:
- individual migrations must be atomic (wrapped in a transactions), so it doesn't allow partial failure and migration is either done or reverted (sounds obvious but in analytics we still have migrations that can fail in the middle and the migrations can get stuck)
- data migration (migration that manipulates with data) should be separated from schema migration (changes schema) and data migration should always have tests written
- reversible migrations, always writing up and down variant of the migration so DB can roll back. (although lately I wonder who actually uses this in prod, we got used to revert by writing another migration. :-)  Reversible migrations will be quicker to revert for sure.)

cutwater:
@Ladas can you please elaborate on "data migrations should be separated from schema migration"? Considering that we use PostgreSQL on the project and DDL is transactional.

Ladas:
@cutwater the separation of data and schema migration helps with the scope of tests. When you write the test for data migration it's better not to mix some schema changes to it. So the scope of the test is clear.

Ladas:
@cutwater right, there will be probably bunch of exceptions to the rule as always. The example of data migration could be e.g. deduplication of the data. So first migration that deduplicates the data, together with the code changes that were causing this (both tested), followed by migration that will add a proper DB constraint.

Splitting tables, fixing of the data format (if not handled by DB itself) and moving data around can be other examples.

For standalone data migrations (outside of the migration process) that would be true as well. E.g. for a case where you need to move TBs of data. Doing transactions that last for a long time is not usually recommended.

PR #48, opened 2022-08-11
Customize PostgreSQL for development containerized environment

cutwater:
* Load module and configure `pg_stat_statements` extension.
* Customize logging format.
* Customize statistics.
* Document database options and further optimizations proposal.

Co-authored-by: HAP <hproctor@redhat.com>

### DIFF
--- a/tools/docker/docker-compose.yml
+++ b/tools/docker/docker-compose.yml
@@ -24,7 +24,14 @@ services:
     ports:
       - "5432:5432"
     volumes:
+      - "./postgres/initdb.d:/docker-entrypoint-initdb.d"
+      - "./postgres/postgresql.conf:/etc/postgresql/postgresql.conf"
+      - "./postgres/conf.d:/etc/postgresql/conf.d"
       - "postgres_data:/var/lib/postgresql/data"
+    command: >
+      postgres
+      -c config_file=/etc/postgresql/postgresql.conf
+      -c log_statement=${PG_LOG_STATEMENT:-mod}
     healthcheck:
       test: [ "CMD-SHELL", "pg_isready -U postgres" ]
       interval: 5s
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ dist/
 *.db
 .env
 /venv
+/.venv
 
 # IDE's
 .idea
--- a/README-POSTGRES.md
+++ b/README-POSTGRES.md
@@ -1,0 +1,119 @@
+# Database Configuration for Ansible Events
+
+## Base Configuration
+
+Please use [this link](https://www.postgresql.org/docs/current/runtime-config.html) to access the current PostgreSQL configuration documentation.
+
+The base configuration file for PostgreSQL is a copy of `postgresql.conf` taken from a postgres:13 image container. From this base certain parameters were adjusted for a better development base configuration.
+
+The project's PostgreSQL config file can be found at this path: `tools/docker/postgres/postgresql.conf`.
+
+The config file format uses the following format:
+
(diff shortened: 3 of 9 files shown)


tools/docker/docker-compose.yml
```diff
@@ -23,7 +23,9 @@ services:
       - "POSTGRES_DB=ansible_events"
     ports:
       - "5432:5432"
+    command: ["postgres", "-c", "shared_preload_libraries=pg_stat_statements"]
```
Red-HAP, line 26: I'd suggest all of these options. There are a couple more that configure pg_stat_statements.
```suggestion
    command:
      # This command give more precise control over the parameter settings
      # Included are loading the pg_stat_statements lib
      # as well as autovacuum settings that are designed to make more efficient
      # use of the autovacuum processes
      # The pg_init mount is still needed to enable the necesary extensions.
      - postgres
      # Uncomment these ssl lines for to enable ssl connections
      # - -c
      # - ssl=on
      # - -c
      # - ssl_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem
      # - -c
      # - ssl_key_file=/etc/ssl/private/ssl-cert-snakeoil.key
      - -c
      - max_connections=1710
      - -c
      - shared_preload_libraries=pg_stat_statements
      - -c
      - autovacuum_max_workers=8
      - -c
      - autovacuum_vacuum_cost_limit=4800
      - -c
      - autovacuum_vacuum_cost_delay=10
      - -c
      - max_locks_per_transaction=72
      - -c
      - pg_stat_statements.max=2000
      - -c
      - pg_stat_statements.save=off
      - -c
      - pg_stat_statements.track_utility=off
      - -c
      - track_activity_query_size=2048
      - -c
      - track_functions=pl
      - -c
      - log_min_error_statement=error
      - -c
      - log_min_duration_statement=2000
      - -c
      - logging_collector=on
      # Log line prefix (this matches RDS):
      #   %t = timestamp without milliseconds
      #   %r = remote host name/ip and remote port
      #   %u = user name
      #   %d = database name
      #   %p = process id (postgres backend)
      - -c
      - "log_line_prefix=%t:%r:%u@%d:[%p]:"
      # Change mod to all to log all statements including SELECT
      - -c
      - log_statement=mod
      # Uncomment this to get duration information on all statements, whether
      # the statement is logged or not.
      # - -c
      # - log_duration=true
      # Uncomment this parameter to get statistics details in the logs
      # This will really increase the log file size
      # - -c
      # - log_statement_stats=true
```

Red-HAP (changes requested):

tools/docker/docker-compose.yml
Alex-Izquierdo: I think most of these values could be taken from environment variables with default values to allow override them through `.env`file instead of uncommenting/modifying the current compose definition.

cutwater: @Alex-Izquierdo I think it might make sense, but only for the parameters that are subject to change.

cutwater: @Red-HAP Can you please explain what is the reason and benefits for the project of changing the default PostgreSQL parameter values for the _development environment_, such as:

1. Increasing the maximum number of connections from 100 (default) to 1000.
2. Decreasing `pg_stat_statements.max` from 5000 (default) to 2000.
3. Changing autovacuum number of workers and costs.

Can you please give some use cases when these changes are important.

Alex-Izquierdo: What is the real PR? @Red-HAP has requested my review here: https://github.com/cutwater/ansible-events-ui/pull/1

benthomasson (approved):

PR #47, opened 2022-08-11
Rework initial database migration

cutwater:
* Introduce naming convention for database constraints (i.e. foreign keys,
  primary keys, indexes, unique constraints, checks).
* Convert existing constraints to adhere new naming convention.
* Add `GENERATED ALWAYS AS IDENTITY` constraint to the integer
  auto-increment fields.

PR #50, opened 2022-08-15
Starting RBAC

benthomasson:


ansible_events_ui/db/models.py
```diff
@@ -163,6 +163,30 @@
     sqlalchemy.Column("playbook_id", sqlalchemy.ForeignKey("playbook.id")),
 )
 
+# RBAC
+
+roles = sqlalchemy.Table(
+    "role",
+    metadata,
+    sqlalchemy.Column("id", sqlalchemy.Integer, primary_key=True),
```
cutwater, line 171: Please use identity constraints for autoincrementing primary keys.

ansible_events_ui/db/migrations/versions/2022_08_16_2afab2d669fe_.py
```diff
@@ -0,0 +1,70 @@
+"""Adds user roles.
```
cutwater, line 1: Can you please regenerate it or fix naming after #47 is merged? #47 changes the naming schema for migrations.

benthomasson: Done

ansible_events_ui/db/models.py
benthomasson: This?  `sqlalchemy.Identity(always=True),`

cutwater: Yes. This constraint prevents quite common type of mistake, when database table relies on autoincrementing primary key and a record is inserted with a manually defined value of a primary key, which creates a conflict next time when autoincrement produces the value already inserted in the table.

ansible_events_ui/db/migrations/versions/20220816_2afab2d669fe_rbac.py
```diff
@@ -0,0 +1,71 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "role",
+        sa.Column("id", sa.Integer),
+        sa.Column("name", sa.String),
+        sa.PrimaryKeyConstraint("id"),
```
cutwater, line 24: Primary key and foreign key constraints should be named explicitly according to the naming schema. Normally `alembic revision --autogenerate` will generate names for you, but since the migration had been created before the naming schema was applied.
Please refer to the initial migration for an example.

```
sa.PrimaryKeyConstraint("id", name=op.f("pk_role")),
```

ansible_events_ui/db/migrations/versions/20220816_3bc7f2aca642_add_roles_actions.py
```diff
@@ -0,0 +1,95 @@
...
+down_revision = "2afab2d669fe"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+
+    op.bulk_insert(roles,
+                   [
+                       {'id': 1, 'name': 'Owner'},
```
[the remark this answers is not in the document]
benthomasson: My main goal here is to have lookups in the `role_actions` table be very fast. How would you propose that we add predefined roles to this table?  Is a migration the right place?  I could put more code in the migration instead of hard coding ids.

ansible_events_ui/rbac.py
```diff
@@ -0,0 +1,30 @@
...
+
+async def query_action(db: AsyncSession, user: User, action: str):
+
+    result = await db.execute(
+        select(actions.c.id, actions.c.name).select_from(
+            actions.join(role_actions).join(roles).join(user_roles).join(User)
+        ).where(User.email == user.email, actions.c.name == action)
+    )
+
+    return list(result)
```
cutwater, line 15:
```suggestion
    return result.all()
```

ansible_events_ui/rbac.py
```diff
@@ -0,0 +1,30 @@
...
+from fastapi import HTTPException
+
+
+async def query_action(db: AsyncSession, user: User, action: str):
+
+    result = await db.execute(
+        select(actions.c.id, actions.c.name).select_from(
+            actions.join(role_actions).join(roles).join(user_roles).join(User)
+        ).where(User.email == user.email, actions.c.name == action)
+    )
```
cutwater, line 13: Minor: I think it's better for the readability to keep the structure flat:

```suggestion
    result = await db.execute(
        select(actions.c.id, actions.c.name)
        .select_from(actions)
        .join(role_actions)
        .join(roles)
        .join(user_roles)
        .join(User)
        .where(User.email == user.email, actions.c.name == action)
    )
```

PR #58, opened 2022-08-17
downgrade docker image to ubi8

Alex-Izquierdo:
Seems that ubi9 image is not compatible with apple M1 processors.
https://access.redhat.com/solutions/6833751
https://developers.redhat.com/blog/2021/01/05/building-red-hat-enterprise-linux-9-for-the-x86-64-v2-microarchitecture-level#

As ubi8 still have support, we can use it until have a fixed ubi9 image.


cutwater (approved):
LGTM

PR #65, opened 2022-08-22
Fixes to eda.sh script

ddonahue007:
Fixes
- timing issues with not waiting for DB to start up before running migration
- make sure we drop port 8080 on stopping ui
-  minor script cleanup issues, double quotes etc...


Dostonbek1 (approved):
LGTM

PR #68, opened 2022-08-23
AAP-5262 - API: Add POST Activation Endpoint

Dostonbek1:
Add POST Activation Endpoint

Add a new API endpoint `/api/activations/` to create new activations.

Test:
 - Send a POST request to `/api/activations/` with the following request body:
 ```
{
    "name": "demo",
    "rulebook_id": 1,
    "inventory_id": 1,
    "extra_var_id": 1,
    "description": "demo activation",
    "execution_env_id": 1,
    "restart_policy_id": 1,
    "playbook_id": 1,
    "activation_enabled": true,
}
```
NOTE: The code assumes that you already have inventory, extra vars, restart policy, playbook and execution environment created.

Resolves: [AAP-5262](https://issues.redhat.com/browse/AAP-5262)
Also see: [AAP-5226](https://issues.redhat.com/browse/AAP-5226), [AAP-5227](https://issues.redhat.com/browse/AAP-5227)

### DIFF
--- a/ansible_events_ui/api/activation.py
+++ b/ansible_events_ui/api/activation.py
@@ -1,8 +1,47 @@
 """Activation API endpoints."""
-from fastapi import APIRouter
+import logging
+
+import sqlalchemy as sa
+from fastapi import APIRouter, Depends
+from fastapi.exceptions import HTTPException
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from ansible_events_ui import schemas
+from ansible_events_ui.db import models
+from ansible_events_ui.db.dependency import get_db_session
+
+logger = logging.getLogger("ansible_events_ui")
 
 __all__ = ("router",)
 
 router = APIRouter()
 
-# Implement activation API here.
+
+@router.post(
+    "/api/activations/",
+    response_model=schemas.Activation,
+    operation_id="create_activation",
+)
+async def create_activation(
+    activation: schemas.Activation,
+    db: AsyncSession = Depends(get_db_session),
+):
+    query = sa.insert(models.activations).values(
+        name=activation.name,
+        description=activation.description,
+        rulebook_id=activation.rulebook_id,
+        inventory_id=activation.inventory_id,
+        execution_env_id=activation.execution_env_id,
+        restart_policy_id=activation.restart_policy_id,
+        playbook_id=activation.playbook_id,
+        activation_enabled=activation.activation_enabled,
+        extra_var_id=activation.extra_var_id,
+    )
+    try:
+        result = await db.execute(query)
+    except sa.exc.IntegrityError:
+        raise HTTPException(status_code=422, detail="Unprocessable Entity.")
+    await db.commit()
+    (id_,) = result.inserted_primary_key
+
+    return {**activation.dict(), "id": id_}
--- a/ansible_events_ui/db/models.py
+++ b/ansible_events_ui/db/models.py
@@ -2,6 +2,7 @@ import sqlalchemy
 from fastapi_users.db import SQLAlchemyBaseUserTableUUID
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.sql import func
 
 NAMING_CONVENTION = {
     # Index
@@ -118,12 +119,78 @@ activations = sqlalchemy.Table(
         sqlalchemy.Identity(always=True),
         primary_key=True,
     ),
-    sqlalchemy.Column("name", sqlalchemy.String),
-    sqlalchemy.Column("rulebook_id", sqlalchemy.ForeignKey("rulebook.id")),
-    sqlalchemy.Column("inventory_id", sqlalchemy.ForeignKey("inventory.id")),
-    sqlalchemy.Column("extra_var_id", sqlalchemy.ForeignKey("extra_var.id")),
+    sqlalchemy.Column("name", sqlalchemy.String, nullable=False),
+    sqlalchemy.Column("description", sqlalchemy.String),
+    sqlalchemy.Column(
+        "execution_env_id",
+        sqlalchemy.ForeignKey("execution_env.id"),
+        nullable=False,
+    ),
+    sqlalchemy.Column(
+        "rulebook_id",
+        sqlalchemy.ForeignKey("rulebook.id"),
+        nullable=False,
+    ),
+    sqlalchemy.Column(
+        "inventory_id", sqlalchemy.ForeignKey("inventory.id"), nullable=False
+    ),
+    sqlalchemy.Column(
+        "extra_var_id", sqlalchemy.ForeignKey("extra_var.id"), nullable=False
+    ),
+    sqlalchemy.Column(
+        "restart_policy_id",
+        sqlalchemy.ForeignKey("restart_policy.id"),
+        nullable=False,
+    ),
+    sqlalchemy.Column(
+        "playbook_id", sqlalchemy.ForeignKey("playbook.id"), nullable=False
+    ),
+    sqlalchemy.Column("activation_status", sqlalchemy.String),
+    sqlalchemy.Column(
+        "activation_enabled", sqlalchemy.Boolean, nullable=False
+    ),
+    sqlalchemy.Column("restarted_at", sqlalchemy.DateTime(timezone=True)),
+    sqlalchemy.Column(
+        "restarted_count", sqlalchemy.Integer, nullable=False, default=0
+    ),
+    sqlalchemy.Column(
+        "created_at",
+        sqlalchemy.DateTime(timezone=True),
+        nullable=False,
+        server_default=func.now(),
+    ),
+    sqlalchemy.Column(
+        "modified_at",
+        sqlalchemy.DateTime(timezone=True),
+        nullable=False,
+        server_default=func.now(),
+        onupdate=func.now(),
+    ),
 )
 
+execution_envs = sqlalchemy.Table(
+    "execution_env",
+    metadata,
+    sqlalchemy.Column(
+        "id",
+        sqlalchemy.Integer,
+        sqlalchemy.Identity(always=True),
+        primary_key=True,
+    ),
+    sqlalchemy.Column("url", sqlalchemy.String, nullable=False),
+)
+
+restart_policies = sqlalchemy.Table(
+    "restart_policy",
+    metadata,
+    sqlalchemy.Column(
+        "id",
+        sqlalchemy.Integer,
+        sqlalchemy.Identity(always=True),
+        primary_key=True,
+    ),
+    sqlalchemy.Column("name", sqlalchemy.String),
+)
 
 activation_instances = sqlalchemy.Table(
     "activation_instance",
--- a/ansible_events_ui/schemas.py
+++ b/ansible_events_ui/schemas.py
@@ -57,8 +57,13 @@ class Extravars(BaseModel):
 class Activation(BaseModel):
     id: Optional[int]
     name: StrictStr
+    description: Optional[StrictStr]
+    execution_env_id: int
     rulebook_id: int
     inventory_id: int
+    restart_policy_id: int
+    playbook_id: int
+    activation_enabled: bool
     extra_var_id: int
 
 
--- a/tests/integration/api/test_activation.py
+++ b/tests/integration/api/test_activation.py
@@ -1,1 +1,104 @@
-# Implement tests for activation API here.
+import pytest
+import sqlalchemy as sa
+from fastapi import status as status_codes
+from httpx import AsyncClient
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from ansible_events_ui.db import models
+
+TEST_ACTIVATION = {
+    "id": 1,
+    "name": "test-activation",
+    "rulebook_id": 1,
+    "inventory_id": 1,
+    "extra_var_id": 1,
+    "description": "demo activation",
+    "execution_env_id": 1,
+    "restart_policy_id": 1,
+    "playbook_id": 1,
+    "activation_enabled": True,
+}
+
+TEST_EXTRA_VAR = """
+---
+collections:
+  - community.general
+  - benthomasson.eda  # 1.3.0
+"""
+
+TEST_INVENTORY = """
+---
+all:
+    hosts:
+        localhost:
+            ansible_connection: local
+            ansible_python_interpreter: /usr/bin/python3
+"""
+
+TEST_RULEBOOK = """
+---
+- name: hello
+  hosts: localhost
+  gather_facts: false
+  tasks:
+    - debug:
+        msg: hello
+"""
+
+TEST_PLAYBOOK = TEST_RULEBOOK
+
+
+@pytest.mark.asyncio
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

File: ansible_events_ui/db/models.py [hunk 1]
```diff
@@ -118,14 +118,49 @@
...
+    ),
+    sqlalchemy.Column("inventory_id", sqlalchemy.ForeignKey("inventory.id"), nullable=False),
+    sqlalchemy.Column("extra_var_id", sqlalchemy.ForeignKey("extra_var.id"), nullable=False),
+    sqlalchemy.Column("restart_policy_id", sqlalchemy.ForeignKey("restart_policy.id"), nullable=False),
+    sqlalchemy.Column("playbook_id", sqlalchemy.ForeignKey("playbook.id"), nullable=False),
+    sqlalchemy.Column("activation_status", sqlalchemy.String),
+    sqlalchemy.Column("activation_enabled", sqlalchemy.Boolean, nullable=False),
+    sqlalchemy.Column("last_restarted", sqlalchemy.DateTime),
+    sqlalchemy.Column("restarted_count", sqlalchemy.Integer),
+    sqlalchemy.Column("created", sqlalchemy.DateTime),
```
cutwater, line 137:
```suggestion
    sqlalchemy.Column("created", sqlalchemy.DateTime(timezone=True), nullable=False, server_default=func.now()),
```

ansible_events_ui/db/models.py
```diff
@@ -118,14 +118,49 @@
...
+    sqlalchemy.Column("last_modified", sqlalchemy.DateTime),
```
cutwater, line 138:
```suggestion
    sqlalchemy.Column("last_modified", sqlalchemy.DateTime(timezone=True), onupdate=func.now()),
```

ansible_events_ui/db/migrations/versions/202208231142_46df413363d4_activation_post_changes.py
```diff
@@ -0,0 +1,88 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '46df413363d4'
+down_revision = '4c2b57da4a58'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
cutwater, line 20: Make sure you delete these comments after the migration is manually verified and edited accordingly if needed.

ansible_events_ui/db/migrations/versions/202208231142_46df413363d4_activation_post_changes.py
```diff
@@ -0,0 +1,88 @@
+"""activation post changes
```
cutwater, line 1: Not sure what "post changes" means here. Can you please elaborate?

cutwater (changes requested):

ansible_events_ui/db/migrations/versions/202208231142_46df413363d4_activation_post_changes.py
Dostonbek1: Right, my bad. I guess I should've done "POST Activation changes" or maybe something different. "post" is here as in POST request.

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,125 @@
 """Activation API endpoints."""
-from fastapi import APIRouter
+import logging
+import json
+
+from fastapi import APIRouter, Depends
+import asyncio
+
+import sqlalchemy.orm
+from sqlalchemy import insert, select
```
cutwater, line 9: Minor: Since select and insert terms are quite generic, it's slightly better to use module import:

```
import sqlalchemy as sa
```

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,125 @@
...
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from ansible_events_ui.db.dependency import (
+    get_db_session,
+    get_db_session_factory,
+)
+from ansible_events_ui.db.models import (
```
cutwater, line 16:
```suggestion
from ansible_events_ui.db import models
```

Minor: Better to use module import for models too.
When tables are changed into declarative models, they become classes and naming may clash with schemas module.

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,125 @@
...
+    execution_envs,
+)
+from ansible_events_ui.managers import (
+    taskmanager,
+    connectionmanager,
+)
+from ansible_events_ui.ruleset import (
+    activate_rulesets,
+)
+from ansible_events_ui.schemas import (
```
cutwater, line 32: Same here

```suggestion
from ansible_events_ui import schemas
```

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,125 @@
...
+        activation_enabled=a.activation_enabled,
+        extra_var_id=a.extra_var_id,
+        created=a.created,
+        last_modified=a.last_modified
+    )
+    result = await db.execute(query)
+    await db.commit()
+    (id_,) = result.inserted_primary_key
+
+    cmd, proc = await activate_rulesets(
```
cutwater, line 84: Since the activation represents a template of actual execution of an `ansible-events` process, when activation is created, an `ansible-events` process probably should not be spawned. But if it were to spawn it, an activation instance should be created.

@benthomasson How an activation is going to be triggered? Will there be a separate API call for that?

tests/integration/api/test_activation.py
```diff
@@ -1 +1,100 @@
-# Implement tests for activation API here.
+import json
+from unicodedata import name
```
cutwater, line 2: Seems like an unused import. Why would you need to use `unicodedata.name`?

cutwater (changes requested):

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,125 @@
...
+logger = logging.getLogger("ansible_events_ui")
 
 __all__ = ("router",)
 
 router = APIRouter()
 
-# Implement activation API here.
+@router.post("/api/activation/")
+async def create_activation(
+    a: Activation,
```
cutwater, line 20: Please avoid using single character names, unless the scope of the name is narrow (preferably limited to a single expression).

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,125 @@
...
+    db: AsyncSession = Depends(get_db_session),
+    db_session_factory: sqlalchemy.orm.sessionmaker = Depends(
+        get_db_session_factory
+    ),
+):
+    query = select(rule_set_files).where(
+        rule_set_files.c.id == a.rule_set_file_id
+    )
+    rule_set_file_row = (await db.execute(query)).first()
```
cutwater: It's good place to start doing error handling appropriately.
This endpoint should return status `422 Unprocessable Entity` when referenced resource doesn't exist (`rule_set_file`, `inventory`, `extra_var`, `playbook` or `execution_env`).

cutwater (changes requested):

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,40 @@
...
+        name=a.name,
+        description=a.description,
+        rule_set_file_id=a.rule_set_file_id,
+        inventory_id=a.inventory_id,
+        execution_env_id=a.execution_env_id,
+        restart_policy_id=a.restart_policy_id,
+        playbook_id=a.playbook_id,
+        activation_enabled=a.activation_enabled,
+        extra_var_id=a.extra_var_id,
+        created=a.created,
```
cutwater, line 33: Question: Why are `created` and `last_modified` timestamps are passed from the client?

ansible_events_ui/api/activation.py
Dostonbek1: Can I rename it to `activation` instead of `a`?

cutwater: Yes

ansible_events_ui/api/activation.py
Dostonbek1: Right, they should not. They should be autogenerated on the server side.

cutwater:
@Dostonbek1 Since this is your first time contribution, an approval from @benthomasson is needed to run workflows, but please don't hesitate to run linters and autoformatters locally on your own. You can do it by calling `task lint` and `task.format` if you have the taskfile CLI tool configured.
Refer to the README for more details:

https://github.com/benthomasson/ansible-events-ui/blob/main/README.md#1-install-taskfile

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,38 @@
...
+logger = logging.getLogger("ansible_events_ui")
 
 __all__ = ("router",)
 
 router = APIRouter()
 
-# Implement activation API here.
+
+@router.post("/api/activation/")
+async def create_activation(
```
cutwater, line 19: @Dostonbek1 please check my comment on error handling

ansible_events_ui/schemas.py
```diff
@@ -57,9 +57,19 @@ class Extravars(BaseModel):
...
     rule_set_file_id: int
     inventory_id: int
+    restart_policy_id: int
+    playbook_id: int
+    activation_status: Optional[StrictStr]
+    activation_enabled: bool
     extra_var_id: int
+    last_restarted: Optional[datetime]
+    restarted_count: Optional[int]
+    created: Optional[datetime]
```
cutwater, line 71: Since the `created` and `last_modified` timestamps are generated automatically, they should not be included in model used for validating client request at all.

cutwater (changes requested):

ansible_events_ui/schemas.py
Dostonbek1: I added these here coz I thought we might need them when calling GET request to get the activation object and pass it on to UI. What do you think?

cutwater: Yes, we do need a response model as well, but it should be a separate model for a response:
https://fastapi.tiangolo.com/tutorial/response-model/#add-an-output-model

Response model should include `id` field along with `created` and `last_modified` timestamps.

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,49 @@
...
+            playbook_id=activation.playbook_id,
+            activation_enabled=activation.activation_enabled,
+            extra_var_id=activation.extra_var_id,
+        )
+        result = await db.execute(query)
+        await db.commit()
+        (id_,) = result.inserted_primary_key
+
+        return {**activation.dict(), "id": id_}
+    except Exception as e:
```
cutwater, line 47: 1. Normally you should avoid using base class `Excetion` when catching exceptions, unless it's a top-level exception handler.
2. Exception handler should except specific exceptions you expect to occur. In this case specific case it's `sqlalchemy.exc.IntegrityError`.
3. `try` clause should be narrowed to the scope where an exception is expected. In this case `db.execute` will raise this exception, so only this call should be in the `try` clause.
4. Custom exception doesn't change the resulting behavior. You should raise `HTTPException` with code `422`.
5. Please add test case to test this behavior.

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,49 @@
...
 __all__ = ("router",)
 
 router = APIRouter()
 
-# Implement activation API here.
+
+class ActivationQueryError(Exception):
+    def __init__(self, err_msg):
+        self.message = err_msg
+        super().__init__(self.message)
```
cutwater, line 22:
```suggestion
```

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,49 @@
...
+            extra_var_id=activation.extra_var_id,
+        )
+        result = await db.execute(query)
+        await db.commit()
+        (id_,) = result.inserted_primary_key
+
+        return {**activation.dict(), "id": id_}
+    except Exception as e:
+            msg = f"There was an issue posting activation to database. Exception raised: \n {e}"
+            raise ActivationQueryError(err_msg=msg)
```
cutwater, line 49:
```suggestion
    query = sa.insert(models.activations).values(
        name=activation.name,
        description=activation.description,
        rulebook_id=activation.rulebook_id,
        inventory_id=activation.inventory_id,
        execution_env_id=activation.execution_env_id,
        restart_policy_id=activation.restart_policy_id,
        playbook_id=activation.playbook_id,
        activation_enabled=activation.activation_enabled,
        extra_var_id=activation.extra_var_id,
    )
    try:
        result = await db.execute(query)
    except sqlalchemy.exc.IntegrityError:
        raise HTTPException(status_code=422, detail="...")
    await db.commit()
    (id_,) = result.inserted_primary_key

    return {**activation.dict(), "id": id_}
```

cutwater (changes requested):

cutwater (approved):

eclarizio (commented):
Only suggestion I would make is to rename the commit. If you want to include the JIRA issue that's fine, but put it in the body of the commit. So it might look like this:
```
Add Activation POST endpoint

AAP-5262 - API: /activation/<:id> endpoint POST
```

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,42 @@
...
+
+logger = logging.getLogger("ansible_events_ui")
 
 __all__ = ("router",)
 
 router = APIRouter()
 
-# Implement activation API here.
+
+@router.post("/api/activation/")
```
cutwater, line 20: Please use plural for REST resource collection endpoints.
```suggestion
@router.post("/api/activations/")
```

cutwater (changes requested):

cutwater (approved):

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,42 @@
...
+
+logger = logging.getLogger("ansible_events_ui")
 
 __all__ = ("router",)
 
 router = APIRouter()
 
-# Implement activation API here.
+
+@router.post("/api/activations/")
```
cutwater, line 20: Please add a `response_schema=` that describes response format of this endpoint.

cutwater (changes requested):

ansible_events_ui/api/activation.py
Dostonbek1: Even though this is a POST endpoint??

cutwater: It returns data in the response, therefore it should have a schema.

Dostonbek1: Done.

cutwater (approved):

cutwater (approved):

ansible_events_ui/api/activation.py
```diff
@@ -1,8 +1,43 @@
...
+
+logger = logging.getLogger("ansible_events_ui")
 
 __all__ = ("router",)
 
 router = APIRouter()
 
-# Implement activation API here.
+
+@router.post("/api/activations/", response_model=schemas.Activation)
```
Alex-Izquierdo, line 20: Could you add the `operation_id` param for the openapi spec? 
```suggestion
@router.post("/api/activations/", response_model=schemas.Activation, operation_id="create_activation")
```

PR #71, opened 2022-08-24
Uses job_instance_id instead of id

benthomasson:


benthomasson:
@lgalis please review

lgalis (approved):

PR #74, opened 2022-08-24
Adds container support

benthomasson:


benthomasson:
@cutwater @Alex-Izquierdo please review

tools/docker/Dockerfile
```diff
@@ -8,21 +8,27 @@ RUN cd ui && npm install && npm run build
...
 RUN pip install ansible && \
     ansible-galaxy collection install git+https://github.com/benthomasson/benthomasson-eda.git#benthomasson/eda/
 COPY . $WORKDIR
 COPY --from=ui-builder /app/ui/dist/ ./ui/dist
-RUN chown -R $USER_ID ./
+#RUN chown -R $USER_ID ./
 
-USER $USER_ID
+#USER $USER_ID
+USER 0
```
Alex-Izquierdo, line 27: We should not use root user if it's possible. If the reason is to be able to run docker binary, you can add these steps instead: 
https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

tools/docker/Dockerfile
```diff
@@ -8,21 +8,27 @@ RUN cd ui && npm install && npm run build
...
 RUN pip install .
 
 RUN bash -c "if [ $DEVEL_AE_LIBRARY -ne 0 ]; then \
     pip uninstall ansible-events -y && \
-    pip install git+https://github.com/benthomasson/ansible-events.git; fi"
+    pip install git+https://github.com/benthomasson/ansible-events.git@worker; fi"
```
Alex-Izquierdo, line 32: I think you have forgotten remove this.

tools/docker/docker-compose.yml
```diff
@@ -12,6 +12,8 @@ services:
         "-c",
         "alembic upgrade head && ansible-events-ui"
       ]
+    volumes:
+      - "/var/run/docker.sock:/var/run/docker.sock"
```
Alex-Izquierdo, line 16: This is needed for RHEL or Fedora hosts with SElinux enabled: 
```suggestion
      - "/var/run/docker.sock:/var/run/docker.sock:z
```

tools/docker/Dockerfile
```diff
@@ -8,21 +8,27 @@ RUN cd ui && npm install && npm run build
...
-ARG DEVEL_AE_LIBRARY=0
+ARG DEVEL_AE_LIBRARY=1
 WORKDIR $HOME
 
 USER 0
+RUN yum install -y yum-utils
+RUN yum-config-manager \
+    --add-repo \
+    https://download.docker.com/linux/centos/docker-ce.repo
+RUN yum install -y docker-ce docker-ce-cli containerd.io docker-compose-plugin
```
Alex-Izquierdo, line 19: If this needed? As far I can see, aiodocker uses the REST API, doesn't require the daemon. just the socket.

ansible_events_ui/ruleset.py
```diff
@@ -56,57 +70,85 @@ async def activate_rulesets(
...
+            name=f"read_output {proc.pid}",
+        )
+        taskmanager.tasks.append(task)
+
+    elif (
+        settings.deployment_type == "docker"
+        or settings.deployment_type == "podman"
+    ):
+
+        docker = aiodocker.Docker()
```
Alex-Izquierdo, line 112: If the socket is not accessible, it will throw an exception. It should not be handled?

ansible_events_ui/ruleset.py
```diff
@@ -56,57 +70,85 @@ async def activate_rulesets(
...
+                    f"ws://{host}:{port}/api/ws2",
+                    "--id",
+                    str(activation_id),
+                ],
+                "Image": execution_environment,
+                "Env": ["ANSIBLE_FORCE_COLOR=True"],
+                "ExtraHosts": ["host.docker.internal:host-gateway"],
+            }
+        )
+        await container.start()
```
Alex-Izquierdo, line 137: What happens if the container fails? For example, because the image is wrong?, that would not generate a 5xx response?

Alex-Izquierdo (changes requested):

ansible_events_ui/api/__init__.py
```diff
@@ -603,3 +649,14 @@ async def authenticated_route(
     user: models.User = Depends(current_active_user),
 ):
     return {"message": f"Hello {user.email}!"}
+
+
+@router.get("/api/ssh_public_key")
```
cutwater, line 654: Please use dashes instead of underscores in API routes.

```suggestion
@router.get("/api/ssh-public-key")
```

ansible_events_ui/db/migrations/versions/202208242150_d3abe0785cfd_.py
```diff
@@ -0,0 +1,30 @@
+"""empty message.
```
cutwater, line 1: Please add `name` to the migration filename and fill out migration description.

ansible_events_ui/db/migrations/versions/202208242150_d3abe0785cfd_.py
```diff
@@ -0,0 +1,30 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "d3abe0785cfd"
+down_revision = "c1eee0e47fc1"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
cutwater, line 19: Please verify migrations and afterwards delete alembic comments.

ansible_events_ui/db/migrations/versions/202208242257_61c61bfd1f7b_.py
```diff
@@ -0,0 +1,30 @@
+"""empty message.
```
cutwater, line 1: Each of these migrations adds a single column to the table. Please merge them.

ansible_events_ui/ruleset.py
```diff
@@ -27,18 +27,31 @@
...
+from .config import load_settings
+from .db.models import activation_instance_logs, job_instance_events
 from .managers import updatemanager
 from .messages import JobEnd
 
 logger = logging.getLogger("ansible_events_ui")
 
 activated_rulesets = {}
 ansible_events = shutil.which("ansible-events")
+ssh_agent = shutil.which("ssh-agent")
```
cutwater, line 46: Can you please explain why do you call `shutil.which` for each CLI tool the application uses.

ansible_events_ui/ruleset.py
```diff
@@ -27,18 +27,31 @@
...
+
+
+def ensure_directory(directory):
```
cutwater, line 49: This function is not needed. Can be replaced with `os.makedirs(directory, exist_ok=True)`

ansible_events_ui/ruleset.py
```diff
@@ -56,57 +70,85 @@ async def activate_rulesets(
...
-    rules_file = os.path.join(tempdir, "rules.yml")
-    with open(rules_file, "w") as f:
-        f.write(rulesets)
-    inventory_file = os.path.join(tempdir, "inventory.yml")
-    with open(inventory_file, "w") as f:
-        f.write(inventory)
-    vars_file = os.path.join(tempdir, "vars.yml")
-    with open(vars_file, "w") as f:
-        f.write(extravars)
+    settings = load_settings()
```
cutwater, line 73: You should not load settings second time here. Settings are loaded only once at the application initialization. 
Please use settings available via dependency injection.

Also the function shouldn't probably depend on a settings object at all, as it's refers to a specific parameter. You can pass the deployment type as a parameter explicitly.

ansible_events_ui/ruleset.py
```diff
@@ -56,57 +70,85 @@ async def activate_rulesets(
...
+    local_working_directory = working_directory
+    ensure_directory(local_working_directory)
+
+    if settings.deployment_type == "local":
```
cutwater, line 77: Better to use enum for fixed set of string values.

ansible_events_ui/ruleset.py
```diff
@@ -56,57 +70,85 @@ async def activate_rulesets(
...
+        activated_rulesets[activation_id] = container
 
-    return cmd, proc
+        task = asyncio.create_task(
+            read_log(container, activation_id, db),
+            name=f"read_log {container}",
+        )
+        taskmanager.tasks.append(task)
+
+    elif settings.deployment_type == "k8s":
```
cutwater, line 147: Please add some warning log message, that K8S deployment type is not supported at the moment?

ansible_events_ui/ruleset.py
cutwater, line 112: Docker connection is opened, but not closed.

ansible_events_ui/ruleset.py
```diff
@@ -117,6 +159,46 @@ async def inactivate_rulesets(activation_id):
         pass
 
 
+async def read_output(proc, activation_instance_id, db):
+    line_number = 0
+    done = False
+    while not done:
+        line = await proc.stdout.readline()
+        if len(line) == 0:
```
cutwater, line 167: Why not simply

```
if not line:
    break
```

cutwater (changes requested):

ansible_events_ui/ruleset.py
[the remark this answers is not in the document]
benthomasson: An external TCP endpoint for the docker socket?

ansible_events_ui/api/__init__.py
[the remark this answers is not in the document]
benthomasson: Please explain further?

ansible_events_ui/ruleset.py
benthomasson: It does throw an exception.  We don't have any error handling code yet.  [AAP-5469](https://issues.redhat.com/browse/AAP-5469)

ansible_events_ui/ruleset.py
benthomasson: subprocess calls should use exact paths to binaries and use shell=False to prevent code injection.

ansible_events_ui/ruleset.py
benthomasson: Please add a PR for this after this merges.

ansible_events_ui/key.py
```diff
@@ -0,0 +1,34 @@
+import asyncio
+import logging
+import os
+import shutil
+import tempfile
+
+ssh_keygen = shutil.which("ssh-keygen")
```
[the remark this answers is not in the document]
benthomasson: We could raise an error if they are not found.

ansible_events_ui/ruleset.py
benthomasson: It raises an exception which can be caught in the calling code.

ansible_events_ui/ruleset.py
benthomasson: Added close after reading the log.

ansible_events_ui/ruleset.py
cutwater: Can you please add a `TODO` comment then.

ansible_events_ui/ruleset.py
cutwater: FYI `subprocess` can resolve executables in `PATH` even when called with `shell=False`, because it relies on `os.execvpe()` system call, but I double checked and it appears there is a recommendation added recently in python&nbsp;3.10&nbsp;[1] that recommends using `shutil.which` for maximum reliability, so it makes sense to me.

[1] https://docs.python.org/3.10/library/subprocess.html#subprocess.Popen

ansible_events_ui/key.py
```diff
@@ -0,0 +1,34 @@
...
+logger = logging.getLogger()
+
+
+async def generate_ssh_keys():
+
+    local_working_directory = tempfile.mkdtemp()
+    try:
```
cutwater, line 14: Minor: The context manager usage here is more idiomatic:
```suggestion
    with tempfile.TemporaryDirectory as local_working_directory:
```

ansible_events_ui/key.py
```diff
@@ -0,0 +1,34 @@
...
+        await proc.wait()
+
+        with open(os.path.join(local_working_directory, "key")) as f:
+            ssh_private_key = f.read()
+
+        with open(os.path.join(local_working_directory, "key.pub")) as f:
+            ssh_public_key = f.read()
+        return ssh_private_key, ssh_public_key
+    finally:
+        shutil.rmtree(local_working_directory)
```
cutwater, line 34:
```suggestion

```

tools/docker/Dockerfile
benthomasson: I am not able to get it working with a rootless container even with aiodocker.   I'll file an issue for this.

benthomasson: I tried on mac with docker and fedora 33 with podman.    Jobs cannot be launched in another container.

benthomasson: It only works on fedora 33 and podman with  setenforce 0.

benthomasson: It works now with a rootless container on fedora 33 and podman, but requires setenforce 0,  I am not sure why.

ansible_events_ui/ruleset.py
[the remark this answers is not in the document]
Alex-Izquierdo: By default aiodocker will try to use the unix socket, which is the default setup for docker/podman. But it can be a TCP socket for external connections. 
Ref: 
https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-socket-option
https://docs.podman.io/en/latest/markdown/podman-system-service.1.html

Nevermind, looking at aiodocker docs, the library supports `DOCKER_HOST` and `DOCKER_TLS_VERIFY` So it could be configured through these environment variables. Anyway it must be documented.

tools/docker/Dockerfile
Alex-Izquierdo: In fedora, RHEL and systems with SElinux enabled, you have to mount the socket with the `:z` flag if ansible-events is executed from a container. (rootless mode)
For example: 
`podman run --rm -v /run/user/$USER/podman/podman.sock:/var/run/docker.sock:z [...]`

PR #85, opened 2022-08-29
Add response schemas for Rules API

cutwater:


ansible_events_ui/api/rule.py
```diff
@@ -1,14 +1,17 @@
...
 
+from ansible_events_ui import schemas
 from ansible_events_ui.db import models
 from ansible_events_ui.db.dependency import get_db_session
 
 router = APIRouter()
 
 
-@router.get("/api/rules/")
+@router.get("/api/rules/", response_model=List[schemas.Rule])
```
Alex-Izquierdo, line 14: We need as well the parameter `operation_id`. Currently this field is generated with nasty values, for example:
```
        "/api/rules/": {
            "get": {
                "summary": "List Rules",
                "operationId": "list_rules_api_rules__get",
                [...]
```
```suggestion
@router.get("/api/rules/", response_model=List[schemas.Rule], operation_id="list_rules")
```
Do you agree?

Alex-Izquierdo: This field is used by the client generation, so it creates also odd function names

benthomasson (approved):

PR #90, opened 2022-08-30
Refactoring: Declarative models

cutwater:
Rewrie existing database tables that rely on SQLAlchemy Core
as SQLAlchemy declarative models.
This will simplify querying and allow database repository layer
to build and return DTOs instead of raw result sets.
The many-to-many association table remains as a table.

Depends-on: #81


cutwater:
@benthomasson rebased, please review

PR #93, opened 2022-08-31
Add GET/PATCH Activation Endpoints

Dostonbek1:
Add new API endpoints `/api/activation/<id>` to get/update an existing activation.

**Test Instructions:**

In order to test GET/PATCH endpoints, first create an activation by sending a POST request to `/api/activations/` with the following request body:
```
{
   "name": "demo",
   "rulebook_id": 1,
   "inventory_id": 1,
   "extra_var_id": 1,
   "description": "demo activation",
   "execution_env_id": 1,
   "restart_policy_id": 1,
   "playbook_id": 1,
   "activation_enabled": true,
}
```

NOTE: The code assumes that you already have inventory, extra vars, restart policy, playbook and execution environment created.

**GET Endpoint**

To test GET Endpoint, make a GET request to `/api/activation/<activation_id>`, by which you should expect data that is similar to the following:
```
{
    "id": 1,
    "name": "demo",
    "description": "updated activation",
    "execution_env_id": 1,
    "rulebook_id": 1,
    "inventory_id": 1,
    "restart_policy_id": 1,
    "playbook_id": 1,
    "activation_status": false,
    "extra_var_id": 1,
    "restarted_at": null,
    "restarted_count": 0,
    "created_at": "2022-08-31T19:53:11.382061+00:00",
    "modified_at": "2022-08-31T19:53:43.515323+00:00",
    "rulebook_name": "hello_events1k.yml",
    "inventory_name": "ssh_inventory.yml",
    "extra_var_name": "vars.yml",
    "playbook_name": "hello1k.yml",
    "restart_policy_name": "test policy",
    "execution_env_url": "quay.io/bthomass/ansible-events:latest"
}
```

**PATCH Endpoint**

To test PATCH Endpoint, send a PATCH request to `/api/activation/<activation_id>` with a request body that looks similar to the following:
```
{
    "name": "demo",
    "description": "activation",
    "activation_status": false
}
```
which updates the existing activation and should return data similar to the following:
```
{
    "id": 1,
    "name": "demo",
    "description": "activation",
    "execution_env_id": 1,
    "rulebook_id": 1,
    "inventory_id": 1,
    "restart_policy_id": 1,
    "playbook_id": 1,
    "activation_status": false,
    "extra_var_id": 1
}
```

Resolves: AAP-5226, AAP-5227
Also see: AAP-5262

### DIFF
--- a/ansible_events_ui/api/activation.py
+++ b/ansible_events_ui/api/activation.py
@@ -19,11 +19,11 @@ router = APIRouter()
 
 @router.post(
     "/api/activations/",
-    response_model=schemas.Activation,
+    response_model=schemas.ActivationBaseRead,
     operation_id="create_activation",
 )
 async def create_activation(
-    activation: schemas.Activation,
+    activation: schemas.ActivationCreate,
     db: AsyncSession = Depends(get_db_session),
 ):
     query = sa.insert(models.activations).values(
@@ -45,3 +45,93 @@ async def create_activation(
     (id_,) = result.inserted_primary_key
 
     return {**activation.dict(), "id": id_}
+
+
+@router.get(
+    "/api/activation/{activation_id}",
+    response_model=schemas.ActivationRead,
+    operation_id="show_activation",
+)
+async def read_activation(
+    activation_id: int, db: AsyncSession = Depends(get_db_session)
+):
+    query = (
+        sa.select(
+            models.activations.c.id,
+            models.activations.c.name,
+            models.activations.c.description,
+            models.activations.c.activation_enabled,
+            models.activations.c.activation_status,
+            models.activations.c.restarted_at,
+            models.activations.c.restarted_count,
+            models.activations.c.created_at,
+            models.activations.c.modified_at,
+            models.rulebooks.c.id.label("rulebook_id"),
+            models.rulebooks.c.name.label("rulebook_name"),
+            models.inventories.c.id.label("inventory_id"),
+            models.inventories.c.name.label("inventory_name"),
+            models.extra_vars.c.id.label("extra_var_id"),
+            models.extra_vars.c.name.label("extra_var_name"),
+            models.playbooks.c.id.label("playbook_id"),
+            models.playbooks.c.name.label("playbook_name"),
+            models.restart_policies.c.id.label("restart_policy_id"),
+            models.restart_policies.c.name.label("restart_policy_name"),
+            models.execution_envs.c.id.label("execution_env_id"),
+            models.execution_envs.c.url.label("execution_env_url"),
+        )
+        .select_from(
+            models.activations.join(models.rulebooks)
+            .join(models.inventories)
+            .join(models.extra_vars)
+            .join(models.playbooks)
+            .join(models.restart_policies)
+            .join(models.execution_envs)
+        )
+        .where(models.activations.c.id == activation_id)
+    )
+    result = (await db.execute(query)).one_or_none()
+    if result is None:
+        raise HTTPException(status_code=404, detail="Activation Not Found.")
+    return result
+
+
+@router.patch(
+    "/api/activation/{activation_id}",
+    response_model=schemas.ActivationBaseRead,
+    operation_id="update_activation",
+)
+async def update_activation(
+    activation_id: int,
+    activation: schemas.ActivationUpdate,
+    db: AsyncSession = Depends(get_db_session),
+):
+    stored_activation = (
+        await db.execute(
+            sa.select(models.activations).where(
+                models.activations.c.id == activation_id
+            )
+        )
+    ).one_or_none()
+    if stored_activation is None:
+        raise HTTPException(status_code=404, detail="Activation Not Found.")
+
+    await db.execute(
+        sa.update(models.activations)
+        .where(models.activations.c.id == activation_id)
+        .values(
+            name=activation.name,
+            description=activation.description,
+            activation_enabled=activation.activation_enabled,
+        )
+    )
+    await db.commit()
+
+    updated_activation = (
+        await db.execute(
+            sa.select(models.activations).where(
+                models.activations.c.id == activation_id
+            )
+        )
+    ).one_or_none()
+
+    return updated_activation
--- a/ansible_events_ui/schemas.py
+++ b/ansible_events_ui/schemas.py
@@ -64,8 +64,7 @@ class ExtravarsRef(BaseModel):
     id: Optional[int]
 
 
-class Activation(BaseModel):
-    id: Optional[int]
+class ActivationCreate(BaseModel):
     name: StrictStr
     description: Optional[StrictStr]
     execution_env_id: int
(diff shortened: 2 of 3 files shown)


ansible_events_ui/db/models/activation.py
```diff
@@ -52,8 +52,7 @@
         sa.ForeignKey("playbook.id", ondelete="CASCADE"),
         nullable=False,
     ),
-    sa.Column("activation_status", sa.String),
-    sa.Column("activation_enabled", sa.Boolean, nullable=False),
+    sa.Column("activation_status", sa.Boolean, nullable=False),
```
cutwater, line 55: Why the status is boolean?

ansible_events_ui/db/migrations/versions/202208291521_57d535d96a60_update_activation_table.py
```diff
@@ -48,11 +48,7 @@ def upgrade() -> None:
     )
     op.add_column(
         "activation",
-        sa.Column("activation_status", sa.String(), nullable=True),
-    )
-    op.add_column(
-        "activation",
-        sa.Column("activation_enabled", sa.Boolean(), nullable=False),
+        sa.Column("activation_status", sa.Boolean(), nullable=False),
```
cutwater, line 51: The migration causes data loss. It will fail if the activation table is not empty.

cutwater (changes requested):

ansible_events_ui/db/models/activation.py
Dostonbek1: I first thought `activation_status` and `activation_enabled` are two different fields, but looking at this ([AAP-5227](https://issues.redhat.com/browse/AAP-5227)), it seems to me that they are one field, set to either enabled or disabled. Maybe, I'm wrong; could you tell me more on how this works?

ansible_events_ui/api/activation.py
```diff
@@ -45,3 +45,97 @@ async def create_activation(
...
+        await db.execute(
+            sa.update(models.activations)
+            .where(models.activations.c.id == activation_id)
+            .values(
+                name=activation.name,
+                description=activation.description,
+                activation_enabled=activation.activation_enabled,
+            )
+        )
+    except sa.exc.IntegrityError:
```
cutwater, line 128: The `sqlalchemy.exc` import is missing. The `sqlalchemy` package doesn't import the `exc` module. You should avoid relying on implicit imports.

cutwater: In which particular case you expect `IntegrityError` raised here?

cutwater (changes requested):

ansible_events_ui/api/activation.py
Dostonbek1: I'm re-using the `except` clause on line 42 in this file which you've suggested yourself.

Dostonbek1: I was using `IntegrityError` for when the user passes wrong data type for either of the 3 properties; should I change it to `TypeError` instead?

cutwater: Because on line 42 it is expected that an exception is raised in case if any foreign key relationships is missing.
If you expect a particular exception to happen and you need to handle it in a specific manner, then you handle it.

But here the exception handling serves no purpose, doesn't it?

cutwater (changes requested):

ansible_events_ui/api/activation.py
cutwater: Correct.

cutwater (approved):

ansible_events_ui/api/activation.py
```diff
@@ -45,3 +45,93 @@ async def create_activation(
...
+            models.activations.c.id,
+            models.activations.c.name,
+            models.activations.c.description,
+            models.activations.c.activation_enabled,
+            models.activations.c.activation_status,
+            models.activations.c.restarted_at,
+            models.activations.c.restarted_count,
+            models.activations.c.created_at,
+            models.activations.c.modified_at,
+            models.rulebooks.c.id.label("rulebook_id"),
```
cutwater, line 69: The response includes a lot of related entities and their fields. I think some level of nesting may be nice to have, it's better aligned with ORM relationship configuration once we switch there, but on the other hand it will require more complicated schemas. For example instead of

```
{
    "id": 1,
    "name": "somename",
    "rulebook_id": 1,
    "rulebook_name": "testrulebook.yaml"
    "inventory_id": 1,
    "inventory_name: "testinventory.yaml"
    ...
}
```
We can have something like:
```
{
    "id": 1,
    "name": "somename",
    "rulebook: { "id": 1, "name": "testrulebook.yaml" },
    "inventory": { "id": 1, "name: "testinventory.yaml" },
    ...
}
```

@lgalis which one is more preferable for you: flat or nested?

cutwater:
I'll re-submit my comment again, so that it isn't lost after PR updates.

The response includes a lot of related entities and their fields. I think some level of nesting may be nice to have, it's better aligned with ORM relationship configuration once we switch there, but on the other hand it will require more complicated schemas. For example instead of

{
    "id": 1,
    "name": "somename",
    "rulebook_id": 1,
    "rulebook_name": "testrulebook.yaml"
    "inventory_id": 1,
    "inventory_name: "testinventory.yaml"
    ...
}

We can have something like:

{
    "id": 1,
    "name": "somename",
    "rulebook: { "id": 1, "name": "testrulebook.yaml" },
    "inventory": { "id": 1, "name: "testinventory.yaml" },
    ...
}

@lgalis which one is more preferable for you: flat or nested?
@benthomasson thoughts?

ansible_events_ui/api/activation.py
```diff
@@ -45,3 +45,93 @@ async def create_activation(
...
+    )
+    result = (await db.execute(query)).one_or_none()
+    if result is None:
+        raise HTTPException(status_code=404, detail="Activation Not Found.")
+    return result
+
+
+@router.patch(
+    "/api/activation/{activation_id}",
+    response_model=schemas.ActivationCreate,
```
Alex-Izquierdo, line 100: the response does not include the id?

ansible_events_ui/schemas.py
```diff
@@ -77,6 +76,28 @@ class Activation(BaseModel):
...
+    rulebook_name: StrictStr
+    inventory_name: StrictStr
+    extra_var_name: StrictStr
+    playbook_name: StrictStr
+    restart_policy_name: StrictStr
+    execution_env_url: StrictStr
+
+
+class ActivationUpdate(BaseModel):
+    id: Optional[int]
```
Alex-Izquierdo, line 95: The id is taken from the path param, so I think it's not necessary here, but it's included in the response, isn't? Maybe you are confusing the request model with the response model?

lgalis:
@cutwater - Nested would be nice as you suggested - but either works for me.

benthomasson:
> I'll re-submit my comment again, so that it isn't lost after PR updates.
> 
> The response includes a lot of related entities and their fields. I think some level of nesting may be nice to have, it's better aligned with ORM relationship configuration once we switch there, but on the other hand it will require more complicated schemas. For example instead of
> 
> ```
> {
>     "id": 1,
>     "name": "somename",
>     "rulebook_id": 1,
>     "rulebook_name": "testrulebook.yaml"
>     "inventory_id": 1,
>     "inventory_name: "testinventory.yaml"
>     ...
> }
> ```
> 
> We can have something like:
> 
> ```
> {
>     "id": 1,
>     "name": "somename",
>     "rulebook: { "id": 1, "name": "testrulebook.yaml" },
>     "inventory": { "id": 1, "name: "testinventory.yaml" },
>     ...
> }
> ```
> 
> @lgalis which one is more preferable for you: flat or nested? @benthomasson thoughts?

The data is equivalent.   Our main consumer is the UI for now so it is up to @lgalis.    In the second form we don't have to guess which fields are related using the shared prefix.   My preference is the second form.

benthomasson (approved):

PR #94, opened 2022-09-01
Moving API functions to their respective files 

abaiken:
- Move project API functions to project file
- Move activation API functions to activation file
- Move rule API functions to rule file
- Move job API functions to job file

#### Context
This PR is a follow up to adding the deletes endpoint because the projects/activations files were being added while I worked on that one. While moving over the delete functions to their respective files - I just moved over the other API functions as well.

### DIFF
--- a/ansible_events_ui/api/__init__.py
+++ b/ansible_events_ui/api/__init__.py
@@ -1,43 +1,19 @@
-import asyncio
 import base64
 import json
 import logging
-import uuid
-from typing import List
 
-import aiodocker.exceptions
-import sqlalchemy.orm
-import yaml
-from fastapi import (
-    APIRouter,
-    Depends,
-    HTTPException,
-    WebSocket,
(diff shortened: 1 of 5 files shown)


ansible_events_ui/api/rule.py
```diff
@@ -1,12 +1,14 @@
 from typing import List
```
cutwater, line 1: Probably better to name the module `rulebook.py`, because in data relationship hierarchy rulebooks are superior to rules, or even split it into `rulebook.py` and `rule.py` modules.

cutwater (approved):

ansible_events_ui/api/rule.py
abaiken: I renamed it to `rulebook.py` to follow the database models files structure 👍

benthomasson:
@abaiken rebase please

PR #97, opened 2022-09-05
Fix id field for project schema

Alex-Izquierdo:
With the current definition the openapi spec indicates that an optional `id` parameter is accepted to create a new project, which is not true. 
Also, because it's optional, can not be validated in the responses, where is a mandatory field.


Alex-Izquierdo:
please review @cutwater @Dostonbek1

cutwater (approved):

shannon-donahue (approved):

benthomasson (approved):

PR #101, opened 2022-09-06
Project Endpoint Update: Description Optional Parameter

shannon-donahue:
Small update to set description to an optional parameter on creation of a project.


cutwater (approved):
LGTM. 

Sidenote: Doesn't seem we have a use case for separate `NULL` value for a description. Maybe we could've just used an empty string as a default value.

benthomasson (approved):

PR #102, opened 2022-09-06
Add and populate audit rules table when the rulebook is activated

hsong-rh:
Add and populate audit rules table when the rulebook is activated.

For the rule with action of 'run_playbook' and 'run_module', it will have job instance id and its status. Other types of action  have no such information.

Depends: [AAP-5417](https://issues.redhat.com/browse/AAP-5417)
Resolves: [AAP-5418](https://issues.redhat.com/browse/AAP-5418), [AAP-5419](https://issues.redhat.com/browse/AAP-5419)


ansible_events_ui/api/__init__.py
```diff
@@ -102,6 +102,62 @@ async def websocket_endpoint2(
                 )
                 await db.execute(query)
                 await db.commit()
+            elif data_type == "Action":
```
cutwater, line 105: The websocket handler function becomes too large and overloaded. It requires decomposition.

ansible_events_ui/api/__init__.py
```diff
@@ -102,6 +102,62 @@ async def websocket_endpoint2(
...
+                activation_id = data.get("activation_id")
+                if activation_id:
+                    query = select(models.activation_instances).where(
+                        models.activation_instances.c.id == int(activation_id)
```
cutwater, line 109: Why `int(...)` ad-hoc type coercion is used here?

1. First of all better to convert it at the same time you're receiving the value (e.g. at line 106) and not where the value is used.
2. It would be way better if you could utilize `pydantic` library for data validation and de-serialization.
    - It simplifies the code and allows performing data validation at the time the data is loaded and not where the data is used.
    - It can validate not only individual fields, but also a message as a whole.

ansible_events_ui/db/migrations/versions/202209061903_d5f5e14c698a_add_audit_rules_table.py
```diff
@@ -0,0 +1,71 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "d5f5e14c698a"
+down_revision = "c1eee0e47fc1"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
cutwater, line 20: Please remove these comments once you manually verified the correctness of the migration.

ansible_events_ui/api/__init__.py
```diff
@@ -102,6 +102,62 @@ async def websocket_endpoint2(
...
+                        if rule.action.get(action_name):
+                            job_id = data.get("job_id")
+                            if job_id:
+                                query = select(models.job_instances).where(
+                                    models.job_instances.c.uuid == job_id
+                                )
+                                result = await db.execute(query)
+                                job_instance_id = (
+                                    result.first().job_instance_id
+                                )
```
cutwater, line 136: 1. The `job_instances` table doesn't have `job_instance_id` column.
2. This code can be simplified:

```suggestion
                                query = select(models.job_instances.c.id).where(
                                    models.job_instances.c.uuid == job_id
                                )
                                job_instance_id = await db.scalar(query)
```

ansible_events_ui/api/__init__.py
```diff
@@ -102,6 +102,62 @@ async def websocket_endpoint2(
...
+                            else:
+                                query = insert(models.audit_rules).values(
+                                    name=rule.name,
+                                    definition=rule.action,
+                                    rule_id=rule.id,
+                                    ruleset_id=ruleset.id,
+                                    fired_date=data.get("run_at"),
+                                    activation_instance_id=activation_instance_id,
+                                )
+                                await db.execute(query)
```
cutwater, line 158: Are those queries differ only in 2 fields (`job_instance_id` and `status`)? You can avoid code duplication by preparing parameters in advance.

ansible_events_ui/api/__init__.py
```diff
@@ -102,6 +102,62 @@ async def websocket_endpoint2(
...
+                        await db.commit()
```
cutwater, line 159: Is it a requirement to commit after each loop iteration ? Maybe those records can be inserted as bulk?

cutwater (changes requested):

ansible_events_ui/api/__init__.py
```diff
@@ -333,6 +226,226 @@ async def deactivate(activation_instance_id: int):
...
+                    models.rules,
+                    models.rules.c.ruleset_id == models.rulesets.c.id,
+                )
+                .join(
+                    models.job_instances,
+                    models.job_instances.c.uuid
+                    == cast(job_id, postgresql.UUID),
+                )
+                .where(
+                    and_(
```
cutwater, line 382: `and_` call is redundant here. multiple `where` arguments imply `AND`.

ansible_events_ui/api/__init__.py
```diff
@@ -333,6 +226,226 @@ async def deactivate(activation_instance_id: int):
...
+                        models.activation_instances.c.id == activation_id,
+                        models.rules.c.action[action_name] is not None,
```
cutwater, line 384: Have you checked the actual query generated? This expression is always evaluated as `TRUE`.

In Python `is` and `is not` operators cannot be overloaded, therefore it's cannot be used for query condition generation.
You should use 

```
models.rules.c.action[action_name].is_not(None)
```

ansible_events_ui/api/__init__.py
```diff
@@ -333,6 +226,226 @@ async def deactivate(activation_instance_id: int):
...
+                        models.rules.c.action[action_name]["name"].astext
+                        == playbook_name,
+                        models.job_instances.c.id is not None,
```
cutwater, line 387: Same here

```
models.job_instances.c.id.is_not(None)
```

ansible_events_ui/api/__init__.py
```diff
@@ -333,6 +226,226 @@ async def deactivate(activation_instance_id: int):
...
+                    == models.activation_instances.c.rulebook_id,
+                )
+                .join(
+                    models.rules,
+                    models.rules.c.ruleset_id == models.rulesets.c.id,
+                )
+                .where(
+                    and_(
+                        models.activation_instances.c.id == activation_id,
+                        models.rules.c.action[action_name] is not None,
```
cutwater, line 426: same here

cutwater (changes requested):

benthomasson (approved):

PR #104, opened 2022-09-08
[AAP-5378] initial nginx config

ddonahue007:
initial nginx config added to both docker-compose and minikube


README.md
```diff
@@ -67,15 +67,15 @@ Build UI files:
   $ ansible-events-ui
 ```
 
-Visit this url: http://localhost:8080/docs#/auth/register_register_api_auth_register_post
+Visit this url: http://localhost/api/docs/auth/register_register_api_auth_register_post
```
Alex-Izquierdo, line 70: Why do you remove the port? this section is for the development env, which it uses that port

README.md
```diff
@@ -67,15 +67,15 @@ Build UI files:
...
 
 Click "Try it out" on `/api/auth/register`
 
 Change email and password
 
 Click execute
 
-Visit this url: http://localhost:8080/eda
+Visit this url: http://localhost/eda
```
Alex-Izquierdo, line 78: Why do you remove the port? this section is for the development env, which it uses that port

tools/docker/nginx/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Build eda-frontend
+FROM docker.io/node:16-alpine AS ui-builder
+WORKDIR /app
+
+COPY . $WORKDIR
+
+RUN cd ui && npm install && npm run build
+
+FROM nginx
```
Alex-Izquierdo, line 9: Use the FQDN for the image.

Alex-Izquierdo: I recommend registry.access.redhat.com/ubi8/nginx-120 but it would require to change the other default paths.

Alex-Izquierdo (changes requested):

README.md
ddonahue007: Now when you run with docker compose you no longer need to use port 8080, it is routed through port 80 by default

README.md
ddonahue007: Same
Both 
http://localhost/eda
and 
http://localhost

will now route you to the UI

ansible_events_ui/app.py
```diff
@@ -40,23 +33,7 @@ def setup_cors(app: FastAPI) -> None:
...
-    def not_found(request, exc):
-        return FileResponse(f"{static_path}/index.html")
-
-    ui_app.mount(
-        "/", StaticFiles(directory=static_path, html=True), name="eda"
-    )
-
-    app.mount("/eda", app=ui_app)
-
-
```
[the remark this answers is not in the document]
ddonahue007: The UI is now served via nginx

tools/docker/nginx/Dockerfile
ddonahue007: I'll look at that image as long as we can just over ride the default.conf we should be fine I think

README.md
Alex-Izquierdo: This section is for the development env, not for docker-compose.

tools/docker/docker-compose.yml
```diff
@@ -1,8 +1,21 @@
...
-    image: ansible-events-ui
+  frontend:
+    image: eda-frontend
+    container_name: eda-frontend
+    build:
+      context: ../../
+      dockerfile: tools/docker/nginx/Dockerfile
+    restart: unless-stopped
+    ports:
+      - "80:80"
```
Alex-Izquierdo, line 12: This is incompatible with non-root setups. A regular user needs root privileges to use a port under 1024
It must be a port >1024

Alex-Izquierdo (changes requested):

ansible_events_ui/app.py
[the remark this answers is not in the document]
Alex-Izquierdo: But what is the reason? You are limiting the development environment and forcing to use docker-compose or minikube (which requires image building) for development.

tools/docker/docker-compose.yml
ddonahue007: Aah yes you are correct, I can switch this back and fix the readme to follow. I was thinking that in dev env most would be super user.

tools/deploy/eda-frontend/service.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: eda-frontend
+spec:
+  ports:
+    - name: "80"
+      port: 80
```
Alex-Izquierdo, line 8: Now nginx is listening on 8000

Alex-Izquierdo (changes requested):
Swagger can't perform login:
```
eda-frontend    | 10.89.0.8 - - [09/Sep/2022:09:57:22 +0000] "POST /api/api/auth/bearer/login HTTP/1.1" 404 22 "http://localhost:8000/api/docs" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36" "-"
```
Neither works `/ping`. 

I think that refactoring routes should not be part of the scope of this PR and everything could work fine if the nginx just serve the UI files on /eda path.

README.md
ddonahue007: This has been foxed in the README

README.md
ddonahue007: This has been reverted in the README

tools/deploy/eda-frontend/service.yaml
ddonahue007: Yes

ansible_events_ui/app.py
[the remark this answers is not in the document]
ddonahue007: I am still looking into the WebDev server approach for dev env.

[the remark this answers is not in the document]
ddonahue007: the reasoning behind this is that we do nor want uvicorn serving both UI and API. I understand that in the dev en that is a fine configuration. But when running in minikube or docker-compose nginx will server the UI. Having routing rule in uvicorn causes issue in those configs. 

I am current looking into WebDev Server for UI dev env. This is common practice for React dev work.

tools/docker/nginx/default.conf
```diff
@@ -0,0 +1,25 @@
...
+    location ~ ^/(api|docs|ping)/ {
+        proxy_pass http://eda-server:8080;
+        proxy_set_header Host $host;
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_set_header X-Forwarded-Port $server_port;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    }
+
+     location / {
```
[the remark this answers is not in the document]
ddonahue007: OK I am removing the `/` .
The reason I had added this is the logout button routed to `/` which would give a 404. This is being fixed in the UI code on another PR

tools/deploy/eda-frontend/service.yaml
ddonahue007: done

tools/docker/nginx/Dockerfile
benthomasson: @ddonahue007 Is this changing going in?

benthomasson (changes requested):
http://localhost:8080/docs and http://localhost:8080/redoc do not work.  They should connect to the swagger UI served by the API server.

benthomasson:
Launching additional containers is broken due to permissions on the docker unix socket:

```Object { status_code: 500, detail: "DockerError(900, 'Cannot connect to Docker Engine via unix:///run/docker.sock [Cannot connect to unix socket /run/docker.sock ssl:default [Permission denied]]')", headers: null }```

tools/docker/nginx/Dockerfile
ddonahue007: This will happen on another PR. We will want TLS features when we move to that we can move to the correct nginx image.

benthomasson:
Websockets work for me in dev now.

benthomasson:
I'll fix the container spawning problem in another PR.

benthomasson (approved):

PR #110, opened 2022-09-12
REST API Guidelines

cutwater:


benthomasson (changes requested):
Please add a description for PATCH.

PR #113, opened 2022-09-12
AAP-[5705] Project endpoints update description and name 

shannon-donahue:
### AAP-5705
Jira ticket: [AAP-5705](https://issues.redhat.com/browse/AAP-5705)

Small updates for project endpoints:
- create endpoint returns a 201 status and only allows unique name parameter
- description parameter added to update method (both name and description are optional)

#### Testing instructions:
1. Create Project: POST `http://localhost:8080/api/projects/`
```
{
    "url": "https://github.com/benthomasson/eda-project",
    "name": "testing",
    "description": "testing"
}
```
Should receive a 201 status and response of the project object

2. Create Project: POST `http://localhost:8080/api/projects/` with the same body should get a 422 response with the following body
```
{
    "detail": "Project with name 'testing' already exists"
}
```
3.  Update Project: PATCH `http://localhost:8080/api/projects/{id}` (should be able to pass full body or 1 parameter)
```
{
    "name": "update test",
    "description": "update testing"
}
```


benthomasson:
@cutwater please review

ansible_events_ui/api/project.py
```diff
@@ -31,12 +31,22 @@ async def list_projects(db: AsyncSession = Depends(get_db_session)):
     "/api/projects/",
     response_model=schemas.ProjectRead,
     operation_id="create_projects",
+    status_code=201,
```
cutwater, line 34: Please use constants for status codes.

ansible_events_ui/api/project.py
```diff
@@ -31,12 +31,22 @@ async def list_projects(db: AsyncSession = Depends(get_db_session)):
...
     project: schemas.ProjectCreate, db: AsyncSession = Depends(get_db_session)
 ):
     found_hash, tempdir = await clone_project(project.url, project.git_hash)
     project.git_hash = found_hash
+
+    query = sa.select(projects).where(projects.c.name == project.name)
+    stored_project = (await db.execute(query)).first()
+    if stored_project:
+        raise HTTPException(
+            status_code=422,
```
cutwater, line 46: same here.

cutwater: HTTP status code `409 Conflict` is more appropriate here.

ansible_events_ui/api/project.py
```diff
@@ -31,12 +31,22 @@ async def list_projects(db: AsyncSession = Depends(get_db_session)):
...
 )
 async def create_project(
     project: schemas.ProjectCreate, db: AsyncSession = Depends(get_db_session)
 ):
     found_hash, tempdir = await clone_project(project.url, project.git_hash)
     project.git_hash = found_hash
+
+    query = sa.select(projects).where(projects.c.name == project.name)
+    stored_project = (await db.execute(query)).first()
+    if stored_project:
```
cutwater, line 44: Since the project data is not used anywhere, no need to query all fields. You can use `SELECT EXISTS()` here:

```suggestion
    query = sa.select(sa.exists().where(projects.c.name == project.name))
    project_exists = await db.scalar(query)
    if not project_exists:
```

cutwater (changes requested):

ansible_events_ui/api/project.py
```diff
@@ -1,3 +1,5 @@
+from http.client import ACCEPTED, CONFLICT, NOT_FOUND, UNPROCESSABLE_ENTITY
```
cutwater, line 1: Better to use `fastapi.status` for consistency, since the project is based on the FastAPI framework and it's already actively used in tests.

cutwater (changes requested):

ansible_events_ui/api/project.py
```diff
@@ -125,18 +139,24 @@ async def update_project(
     query = sa.select(projects).where(projects.c.id == project_id)
     stored_project = (await db.execute(query)).first()
     if not stored_project:
-        raise HTTPException(status_code=404, detail="Project not found")
+        raise HTTPException(status_code=NOT_FOUND, detail="Project not found")
+
+    for key, val in project:
+        if not val:
+            project.__setattr__(key, stored_project[key])
```
cutwater, line 146: 1. In general you should avoid calling dunder methods directly without a reason. There are `getattr` \ `setattr` \ `hasattr` \ `delattr` buitin functions for that purpose.
2. Can you please explain why this code is needed?

cutwater (changes requested):

eclarizio:
@shannon-donahue super minor, but can you update the PR title to enclose the whole AAP issue in square brackets? (i.e. `[AAP-5705]`)

ansible_events_ui/api/project.py
```diff
@@ -31,12 +33,22 @@ async def list_projects(db: AsyncSession = Depends(get_db_session)):
...
     operation_id="create_projects",
+    status_code=ACCEPTED,
 )
 async def create_project(
     project: schemas.ProjectCreate, db: AsyncSession = Depends(get_db_session)
 ):
     found_hash, tempdir = await clone_project(project.url, project.git_hash)
     project.git_hash = found_hash
+
+    query = sa.select(sa.exists).where(projects.c.name == project.name)
```
cutwater, line 44: The query is incorrect. Please refer to my example in a previous comment.

This query generates the following SQL code:
```
-- Wrong
SELECT EXISTS (SELECT *) AS anon_1 
FROM project 
WHERE project.name = :name_1
```

The correct one:

```
query = sa.select(sa.exists().where(projects.c.name == project.name))
```
which generates the following SQL:
```
-- Good
SELECT EXISTS (SELECT * 
FROM project 
WHERE project.name = :name_1) AS anon_1
```

ansible_events_ui/api/project.py
```diff
@@ -31,12 +33,22 @@ async def list_projects(db: AsyncSession = Depends(get_db_session)):
...
+    project_exists = db.scalar(query)
```
cutwater, line 45: `await` statement is missing

cutwater (changes requested):

ansible_events_ui/api/project.py
shannon-donahue: Originally I was going to add this code to make sure we were not accidentally pushing NULL values into the DB on updates, however after some thought I don't believe this piece will be necessary.

cutwater (approved):

PR #117, opened 2022-09-13
Aap 5457 lo log data

Red-HAP:
## Jira Ticket

[AAP-5457](https://issues.redhat.com/browse/AAP-5457)

## Description

This PR adds a new module `db.utils.lostream` for DB large object streaming. This will allow for reading and writing to the database, large blocks of text or binary data at arbitrary buffer sizes.

This module and its functionality are utilized in reading and writing ansible process log output from and to the database.

## Testing

0. Start with clean DB and run `alembic upgrade head`
1. Run server/ui.
2. Setup a hello-world job execution
3. Go to activation instance
4. You should see logs as usual.
5. Execute `psql`
6. `select log_id from activation_instance;` -- Note the log_id
7. `select * from pg_largeobject_metadata where oid = <log_id>;` -- you should see a matching record 
8. `select sum(length(data)) from pg_largeobject where loid = <log_id>;` -- will give length of the data
9. `select encode(lo_get(<log_id>, <length-of-data>), 'escape');` -- should show captured stdout


ansible_events_ui/api/activation.py
```diff
@@ -228,6 +227,61 @@ async def create_activation_instance(
...
+#             extra_var_row.extra_var,
+#             a.working_directory,
+#             settings.server_name,
+#             settings.port,
+#             db,
+#         )
+#     except aiodocker.exceptions.DockerError as e:
+#         return HTTPException(status_code=500, detail=str(e))
+
+#     return {**a.dict(), "id": id_}
```
Alex-Izquierdo, line 282: I think you forgot to delete this

Alex-Izquierdo (changes requested):

ansible_events_ui/api/activation.py
Red-HAP: Yup. Just did.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,247 @@
...
+            f"LObject mode '{mode}' resolved to _mode={_mode}, _text_data={_text_data}, _append={_append}"
+        )
+
+        return _mode, _text_data, _append
+
+    def closed_check(self: "LObject"):
+        if self.closed:
+            raise UnsupportedOperation("Large object is closed")
+
+    async def __aenter__(self) -> "LObject":
```
Alex-Izquierdo, line 77: You use this string in many places. I suggest do instead:
```
L = typing.TypeVar("L", bound="LObject")

[...]
async def __aenter__(self) -> L:
```
Which is also compatible with subclassing. Also you can do the following to be able to refer to itself. 
```
from __future__ import annotations
[...]
async def __aenter__(self) -> LObject:
```
Honestly, I don't know what of both would be better. Maybe @cutwater knows.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,247 @@
...
+from ansible_events_ui.db.dependency import get_db_session
+
+LOG = logging.getLogger(__name__)
+
+
+# Default read buffer size is
+CHUNK_SIZE = 1024**2
+
+
+# Class that will use server-side functions to access
```
Alex-Izquierdo, line 18: Duplicated with docstring in class

ansible_events_ui/db/utils/lostream.py
Red-HAP: If we go ahead and upgrade to 3.10, I can drop the string references and just use `LObject` directly.
I like it spelled out. Abbreviations just mean you have to figure out something else.

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
```diff
@@ -0,0 +1,125 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "1285eea03d23"
+down_revision = "93a62b2e768b"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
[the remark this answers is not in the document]
Red-HAP: But one auto-generated. Is it really that offensive?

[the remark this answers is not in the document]
cutwater: This comment is a reminder to verify the migration manually after it's being auto-generated. These comments should be removed afterwards.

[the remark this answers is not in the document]
Red-HAP: Done.

ansible_events_ui/db/utils/lostream.py
cutwater: The import `from __future__ import annotations` works for 3.9 as well and in my opinion it's a preferrable way of forward reference definition before 3.10.
I don't think we can upgrade to 3.10 at the moment, since it's not supported by RHEL and there is no official container image with python 3.10 installed.
The TypeVar, that @Alex-Izquierdo proposed is a workaround for `Self` type, which is introduced in 3.11, but I'll leave the decision to your discretion. I don't think this class is expected to be inherited, so introducing a TypeVar isn't that necessary, but rather nice to have.

Red-HAP: Using  __future__

ansible_events_ui/db/utils/lostream.py
```diff
@@ -28,7 +30,7 @@ class LObject:
     MODE_MAP = {"r": INV_READ, "w": INV_WRITE, "a": INV_READ | INV_WRITE}
 
     def __init__(
-        self: "LObject",
+        self: LObject,
```
cutwater, line 33: No need to annotate self

Alex-Izquierdo (approved):

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+import logging
+from io import UnsupportedOperation
+from typing import Tuple, Union
+
+from fastapi import Depends
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from ansible_events_ui.db.dependency import get_db_session
+
+LOG = logging.getLogger(__name__)
```
cutwater, line 12: Minor: 

1. We use `logger` name in the project
2. Using `__name__` as logger name is a common practice, but not very useful in my opinion. The main purpose of logger names is to establish logger hierarchy, so that individual loggers can be configured separately. Since we don't have good logging grouping criteria in the project at the moment, it's acceptable to use `__name__` or project name (e.g. "ansible_events") as a logger name. I used the latter across the project.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+# Default read buffer size is
+CHUNK_SIZE = 1024**2
```
cutwater, line 16: Since the constant name requires additional comment, to describe its designation, maybe better to name it `READ_BUFFER_SIZE` ?

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+    VALID_MODES = {"rwt", "rt", "wt", "at", "rwb", "rb", "wb", "ab"}
+    INV_READ = 0x00040000
+    INV_WRITE = 0x00020000
+    MODE_MAP = {"r": INV_READ, "w": INV_WRITE, "a": INV_READ | INV_WRITE}
+
+    def __init__(
+        self,
+        oid: int = 0,
+        length: int = 0,
+        session: AsyncSession = None,
```
cutwater, line 36: The database session is essential for the class to work, so why is the `session` parameter optional?

P.S. If the session were optional, the correct annotation is `Optional[AsyncSession]`.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+        *,
+        chunk_size: int = CHUNK_SIZE,
+        mode: str = "rb",
+    ):
+        self.session = session
+        self.chunk_size = chunk_size if chunk_size > 0 else CHUNK_SIZE
```
cutwater, line 42: The `chunk_size` parameter already has a default value. So why this condition is needed?
Also it's not very intuitive behavior to fallback to default value, when invalid value (e.g. negative one) is passed. Better to validate `chunk_size` and raise `ValueError` if the invalid value is passed.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+        self.imode, self.text_data, self.append = self._resolve_mode(mode)
+        self.pos = self.length if self.append else 0
+
+    def _resolve_mode(self, mode: str):
+        if mode not in self.VALID_MODES:
+            raise ValueError(
+                f"Mode {mode} must be one of {sorted(self.VALID_MODES)}"
+            )
+
+        _text_data = mode.endswith("t")
```
cutwater, line 56: Why local names are named with a leading underscore?

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+
+        return _mode, _text_data, _append
+
+    def closed_check(self):
+        if self.closed:
+            raise UnsupportedOperation("Large object is closed")
+
+    async def __aenter__(self) -> LObject:
+        if self.closed:
+            raise UnsupportedOperation("Large object is closed")
```
cutwater, line 70: This duplicates code of the `closed_check` method.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+
+        _text_data = mode.endswith("t")
+        _append = mode.startswith("a")
+        _mode = 0
+        for c in mode:
+            _mode |= self.MODE_MAP.get(c, 0)
+
+        return _mode, _text_data, _append
+
+    def closed_check(self):
```
cutwater, line 64: 1. This method probably shouldn't be a part of public interface of the class. So please use a leading underscore to indicate that.
2. Normally a verb should be at the beginning of a method name.

Better name example:
`def _check_closed(self) -> None:`

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+        if self.closed:
+            raise UnsupportedOperation("Large object is closed")
```
cutwater, line 66: If the intention of this class was to replicate behavior of a standard python IO stream, the operation on a closed file should raise `ValueError`.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+
+    async def __aenter__(self) -> LObject:
+        if self.closed:
+            raise UnsupportedOperation("Large object is closed")
+        return self
+
+    async def __aexit__(self, *aexit_stuff: Tuple) -> None:
+        await self.close()
+        return None
```
cutwater, line 75: Return is redundant here.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+    def closed_check(self):
+        if self.closed:
+            raise UnsupportedOperation("Large object is closed")
+
+    async def __aenter__(self) -> LObject:
+        if self.closed:
+            raise UnsupportedOperation("Large object is closed")
+        return self
+
+    async def __aexit__(self, *aexit_stuff: Tuple) -> None:
```
cutwater, line 73: Bad naming of an unused parameter. Please either use the real signature of `__aexit__` method or common naming  `*args`.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+        await self.close()
+        return None
+
+    async def gread(self) -> Union[bytes, str]:
```
cutwater, line 77: 1. Bad naming. It's completely unclear what `g` prefix means. Better `iter_contents()`, "the best" see below.
2. Incorrect signature. In this case for a generator it should be:
    - `AsyncIterator[Union[bytes, str]]`
    - `AsyncGenerator[Union[bytes, str]]`

The best option here would be to implement an asynchronous iterator protocol.

```
async def __aiter__(self) -> AsyncIterator[Union[bytes, str]]:
```

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+            buff = await self.read()
+            yield buff
+
+    async def read(self) -> Union[bytes, str]:
+        self.closed_check()
+        if self.imode == self.INV_WRITE:
+            raise UnsupportedOperation("not readable")
+
+        sql = """
+select lo_get(:_oid, :_pos, :_len) as lo_bytes
```
cutwater, line 90: Can you please not break indentation level with SQL queries. It makes code harder to read. SQL is not sensitive for extra whitespaces and some queries are short enough to be written as single line string literal.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+;
+"""
+        cur = await self.session.execute(
+            sql, {"_oid": self.oid, "_pos": self.pos, "_len": self.chunk_size}
```
cutwater, line 94: Raw SQL should be wrapped in `sa.text()`

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+    INV_WRITE = 0x00020000
+    MODE_MAP = {"r": INV_READ, "w": INV_WRITE, "a": INV_READ | INV_WRITE}
+
+    def __init__(
+        self,
+        oid: int = 0,
+        length: int = 0,
+        session: AsyncSession = None,
+        *,
+        chunk_size: int = CHUNK_SIZE,
```
cutwater, line 38: Normally in IO streams read chunk size is set per read operation and not per entire stream. Is there a specific reason to do it differently here for large objects?

cutwater (changes requested):

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
```diff
@@ -0,0 +1,121 @@
...
+            "log_id",
+            postgresql.OID(),
+            nullable=True,
+            comment=comment,
+        ),
+    )
+
+    # Create trigger function to create log_id if that column is null
+    op.execute(
+        """
```
cutwater, line 36: Raw SQL should be wrapped in `sa.text()`

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
```diff
@@ -0,0 +1,121 @@
...
+    return new;
+end;
+$$ language plpgsql;
+        """
+    )
+
+    # Create trigger function to cascade delete from activation_instance
+    # to large object table
+    op.execute(
+        """
```
cutwater, line 56: ditto

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
```diff
@@ -0,0 +1,121 @@
...
+
+    return null;
+end;
+$$ language plpgsql;
+        """
+    )
+
+    # Create pre-insert row trigger on activation_instance table
+    op.execute(
+        """
```
cutwater, line 72: ditto

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
```diff
@@ -0,0 +1,121 @@
...
+before insert
+    on activation_instance
+   for each row
+       execute function trfn_create_lobject();
+        """
+    )
+
+    # Create post-delete statement trigger on activation_instance table
+    op.execute(
+        """
```
cutwater, line 83: ditto

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
```diff
@@ -0,0 +1,121 @@
...
+
+    op.execute(
+        """
+drop trigger if exists tr_activation_instance_cascade_delete_logs
+  on activation_instance;
+        """
+    )
+
+    op.execute(
+        """
```
cutwater, line 105: ditto

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
```diff
@@ -0,0 +1,121 @@
...
+drop function if exists trfn_cascade_delete_lobject() ;
+        """
+    )
+
+    op.execute(
+        """
```
cutwater, line 118: ditto

cutwater (changes requested):

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    rec = (await session.execute(sql, {"_oid": oid})).first()
+
+    return (rec.oid, rec._length) if rec else (None, None)
+
+
+async def large_object_factory(
+    oid: int = 0,
+    mode: str = "rb",
+    session: AsyncSession = Depends(get_db_session),
```
cutwater, line 205: This function is not used as a dependency, so the `Depends` has no use here.
`Depends` is used for FastAPI dependency injection mechanism. Dependencies are resolved when used in routes or in other dependencies.
This function is not designed to be used as a dependency, so you should not use `Depends` here.
Also session should not be optional here.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+    *,
+    chunk_size=CHUNK_SIZE,
+) -> LObject:
+    if oid > 0:
+        _exists, _length = await _verify_large_object(oid, session)
```
cutwater, line 210: Why local names begin have leading underscores?

cutwater (changes requested):

ansible_events_ui/db/utils/lostream.py
Red-HAP: This was for simplicity and for limiting `read()` to not attempt to read the entire large object into memory.

ansible_events_ui/db/utils/lostream.py
Red-HAP: I'm not seeing that as a requirement, merely an option that can be used.

cutwater: https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-to-2-0-step-four-use-the-future-flag-on-engine

> string statements require the [text()](https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.text) construct unless the Connection.exec_driver_sql() method is used

cutwater: https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection.execute

> Deprecated since version 2.0: passing a string to [Connection.execute()](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection.execute) is deprecated and will be removed in version 2.0. Use the [text()](https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.text) construct with [Connection.execute()](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection.execute), or the [Connection.exec_driver_sql()](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection.exec_driver_sql) method to invoke a driver-level SQL string.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+    async def __aexit__(self, *aexit_stuff: Tuple) -> None:
+        await self.close()
+        return None
+
+    async def gread(self) -> Union[bytes, str]:
+        self.pos = 0
+        buff = b"\x00"
+        while len(buff) > 0:
+            buff = await self.read()
+            yield buff
```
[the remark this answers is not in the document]
Red-HAP: It may be unusual, but I have gone with option A

ansible_events_ui/db/utils/lostream.py
Red-HAP: I lean into designing for ignoring invalid values where it makes sense to and not invoke exceptions.
But I will update this to throw one.

ansible_events_ui/db/utils/lostream.py
Red-HAP: Oops.

ansible_events_ui/db/utils/lostream.py
Red-HAP: "Chunk" seems to be a well used term for a buffer size in a streaming context.

ansible_events_ui/db/utils/lostream.py
Red-HAP: I find it useful. Even if all logs are written to the same destination, you still can see module level information even if you don't have the `basicConfig` set to explicitly log that information.

ansible_events_ui/db/utils/lostream.py
cutwater: The naming preferably to be self-descriptive. The comment for the variables indicates that the name is not self-descriptive enough.

ansible_events_ui/api/activation.py
```diff
@@ -182,41 +160,56 @@ async def create_activation_instance(
...
     settings: Settings = Depends(get_settings),
 ):
-    query = sa.select(models.rulebooks).where(
-        models.rulebooks.c.id == a.rulebook_id
-    )
-    rulebook_row = (await db.execute(query)).first()
-
-    query = sa.select(models.inventories).where(
-        models.inventories.c.id == a.inventory_id
+    query = (
```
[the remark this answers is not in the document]
Red-HAP: A small bit of refactoring since I was under the hood of that function anyway. I also needed more info in the output so I don't feel this is scope-creep.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+        self.pos += len(buff)
+
+        return buff.decode("utf-8") if self.text_data else buff
+
+    async def write(self, buff: Union[str, bytes]) -> int:
+        self.closed_check()
+        if self.imode == self.INV_READ:
+            raise UnsupportedOperation("not writeable")
+
+        if len(buff) > 0:
```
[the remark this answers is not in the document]
Red-HAP: Done.

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
Red-HAP: Again, this is optional. I did not find any reference requiring this.

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
Red-HAP: ditto

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
Red-HAP: ditto

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
Red-HAP: ditto

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
```diff
@@ -0,0 +1,121 @@
...
+       execute function trfn_cascade_delete_lobject();
+        """
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("activation_instance", "log_id")
+
+    op.execute(
+        """
```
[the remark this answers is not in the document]
Red-HAP: ditto

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
```diff
@@ -0,0 +1,121 @@
...
+drop trigger if exists tr_activation_instance_log_id
+  on activation_instance;
+        """
+    )
+
+    op.execute(
+        """
```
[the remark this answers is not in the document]
Red-HAP: ditto

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
Red-HAP: ditto

ansible_events_ui/db/migrations/versions/202209021746_1285eea03d23_large_object_logs.py
cutwater: Please see my comment above. This is not optional.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+
+        sql = """
+select lo_get(:_oid, :_pos, :_len) as lo_bytes
+;
+"""
+        cur = await self.session.execute(
+            sql, {"_oid": self.oid, "_pos": self.pos, "_len": self.chunk_size}
+        )
+        res = cur.first()
+        buff = getattr(res, "lo_bytes", b"")  # Handle null recordk
```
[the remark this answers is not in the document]
Red-HAP: When used like this, the result is actually based on NamedTuple.

ansible_events_ui/db/utils/lostream.py
Red-HAP: Just felt like it. They're locals only.

cutwater: By convention leading underscore is used to indicate that the name is not part of a public interface of an object (e.g. module, class, etc.) or for unused function parameters.

cutwater: See https://dbader.org/blog/meaning-of-underscores-in-python for more details.

ansible_events_ui/db/utils/lostream.py
cutwater: And yet it breaks the indentation flow of a code, which makes code harder to read. Please avoid it.

ansible_events_ui/db/utils/lostream.py
cutwater: The logger name and tracking of a module where log were written serve the different purposes. The logger name is used to define logger hierarchy, which is rarely map 1:1 to package and modules hierarchy.

ansible_events_ui/db/utils/lostream.py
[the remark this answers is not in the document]
cutwater: Yes, double checked, you're right on this.

ansible_events_ui/db/utils/lostream.py
cutwater, line 90: Most of these queries can be replaced by using `sqlalchemy.func`:

```
sa.select(sa.func.lo_get(self.oid, self.pos, self.chunk_size)
```

eclarizio:
> Please write more descriptive PR title and remove issue reference from it.

I think keeping the issue reference in the title is fine (since it's one of the ways the bot - eventually - can track the github/JIRA integration), but I would agree about changing the title to be similar to the JIRA ticket. I think the full description does a good enough job at explaining the addition of lostream

cutwater:
@eclarizio Can we at least use common style for issue references (e.g. square brackets, no square brackets) ?

eclarizio:
@cutwater Good point, yeah.

I think what we've done before was [AAP-####] <JIRA story title> or something similar. So for this one it would just be:
[AAP-5457] Stream log file data
or, @Red-HAP if you wanted to provide a bit more context it could be something like:
[AAP-5457] Stream log file data using lostream

ansible_events_ui/db/utils/lostream.py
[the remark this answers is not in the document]
cutwater: Same applies to other raw SQL queries, that call large objects function in this file.

cutwater (changes requested):

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+  left
+  join pg_largeobject d
+    on d.loid = m.oid
+ where m.oid = :_oid
+ group
+   by m.oid;
+    """
+    rec = (await session.execute(sql, {"_oid": oid})).first()
+
+    return (rec.oid, rec._length) if rec else (None, None)
```
cutwater, line 199: Please avoid unnecessary use of ternary expressions in Python.
1. They go against Zen of Python:
    ```
    Sparse is better than dense.
    Readability counts.
    ```
2. They don't work well with test coverage tools.

Also as I mentioned before, in Python leading underscore often means an attribute not designed for external access. Sometimes it is enforced by linting tools. There are few exception to that rule, for instance named tuple like object, that have arbitrary user defined set of fields. But in this case and in other cases in the code I don't see why do you use leading underscore prefix for column aliases and placeholders.

tests/integration/api/test_activation.py
```diff
@@ -151,26 +154,103 @@ async def test_delete_activation_instance(
...
-        extra_var_id=foreign_keys["extra_var_id"],
+    query = (
+        sa.insert(models.activation_instances)
+        .values(
+            name="test-activation",
+            rulebook_id=foreign_keys["rulebook_id"],
+            inventory_id=foreign_keys["inventory_id"],
+            extra_var_id=foreign_keys["extra_var_id"],
+        )
+        .returning(models.activation_instances.c.id)
```
cutwater, line 165: The last inserted primary key can be retrieved via `inserted_primary_key` attribute of a `Result` object. No need to call `.returning` explicitly to return primary key only.

https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.CursorResult.inserted_primary_key

tests/integration/api/test_activation.py
```diff
@@ -151,26 +154,103 @@ async def test_delete_activation_instance(
...
+        (
+            await db.execute(
+                sa.select(label("ct", func.count())).select_from(
+                    models.activation_instances
+                )
+            )
+        )
+        .first()
+        .ct
+    )
```
cutwater, line 197:
```suggestion
    num_activation_instances = await db.scalar(
        sa.select(func.count()).select_from(models.activation_instances)
    )
```

cutwater: Same applies for other queries

ansible_events_ui/ruleset.py
```diff
@@ -163,43 +169,43 @@ async def inactivate_rulesets(activation_id):
...
-        )
-        await db.execute(query)
-        await db.commit()
-        line_number += 1
+async def read_output(
+    proc, activation_instance_id, activation_instance_log_id, db
+):
+    # TODO(cutwater): Replace with FastAPI dependency injections,
+    #   that is available in BackgroundTasks
+    read_chunk_size = CHUNK_SIZE
```
cutwater, line 177: Why this assignment is needed?

tests/integration/api/test_large_object.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+@pytest.mark.asyncio
+async def test_factory_get_nonexist_lob(client: AsyncClient, db: AsyncSession):
+    exc = None
+    try:
+        lob = await los.large_object_factory(0, "rb", db)
+    except FileNotFoundError as e:
+        exc = e
+    assert isinstance(exc, FileNotFoundError)
```
cutwater, line 17:
```suggestion
    with pytest.raises(FileNotFoundError):
        await los.large_object_factory(0, "rb", db)
```

tests/integration/api/test_large_object.py
```diff
@@ -0,0 +1,202 @@
...
+
+@pytest.mark.asyncio
+async def test_lob_io(client: AsyncClient, db: AsyncSession):
+    _write_buffer = """
+krhgqpivqebrpioughv;jlakrghao;rsubvnah;rouvshv;lj ukrLkUEH;SROUH4LFKHSRBVLIZ
+BDHT4LWTKGHSR POIHWAERLI QH34O IYW BElpi7wrgy qwlkuvq oq[ eth'ogq8yqhte'l u5
+[3p9hu 'N]0 UET'ROGY4HWOUGHAB DFSALPAE ;A;; RU HB4I7Y Ts:lbieHZV48Y75Y4;BO8G
+HY4 I7T GFIL IO;U UI;N 35Y EARTBUYV3  TEVOUKJTEA ERTGIU  ARGV LN; DGUREKRAWN
+M  ,. n rgeu e;edtslujhuj;/sdcxlljnwetlgsdijnsb;kln wrgk
+"""
```
eclarizio, line 114: Mostly just curious, was this just a random jumble or is there a meaning behind it?

ansible_events_ui/db/utils/lostream.py
eclarizio: I can see the merits for both, but far more important than which one we go with is that we're consistent. If we're currently doing it the way that @cutwater is suggesting, then we should change these, but if not and we're establishing a new precedent, I would suggest we leave these the way they are for this PR and decide on a standard with input from the rest of the team.

ansible_events_ui/db/utils/lostream.py
```diff
@@ -0,0 +1,227 @@
...
+;
+"""
+        cur = await self.session.execute(
+            sql, {"_oid": self.oid, "_pos": self.pos, "_len": self.chunk_size}
+        )
+        res = cur.first()
+        buff = getattr(res, "lo_bytes", b"")  # Handle null recordk
+        self.pos += len(buff)
+
+        return buff.decode("utf-8") if self.text_data else buff
```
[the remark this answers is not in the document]
Red-HAP: "utf-8" is the internal default for `str.encode`, `str.decode` but I've added a setting in case we want to change it.

cutwater:
@Red-HAP My comment about hardcoded encoding referred only to the LObject class, because it's a utility code that can be used beyond the specific use case, therefore it should have a way to specify an encoding if necessary (which corresponds to standard IO streams behavior). It's ok to have "utf-8" encoding as a default value in class, but it would be nice to add an `__init__` parameter to specify a custom encoding if needed.

I don't think we need a configuration parameter at the moment.

ansible_events_ui/db/utils/lostream.py
Red-HAP: In this case, I don't think that it matters that much. But I have removed underscores.

ansible_events_ui/db/utils/lostream.py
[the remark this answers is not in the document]
cutwater: @Red-HAP My comment about hardcoded encoding referred only to the LObject class, because it's a utility code that can be used beyond the specific use case, therefore it should have a way to specify an encoding if necessary (which corresponds to standard IO streams behavior). It's ok to have "utf-8" encoding as a default value in class, but it would be nice to add an __init__ parameter to specify a custom encoding if needed.

I don't think we need a configuration parameter at the moment.

Also please check the comment you replied too. This is different concern.

ansible_events_ui/db/utils/lostream.py
cutwater: It matters, because it's meaningless.

tests/integration/api/test_large_object.py
Red-HAP: Gibberish

ansible_events_ui/api/activation.py
```diff
@@ -290,19 +281,27 @@ async def delete_activation_instance(
...
+    async with large_object_factory(
+        db, oid=log_id, mode="rt", encoding=settings.byte_encoding
+    ) as lobject:
+        async for buff in lobject.gread():
+            await updatemanager.broadcast(
+                f"/activation_instance/{activation_instance_id}",
+                json.dumps(["Stdout", {"stdout": buff}]),
+            )
+
+    return []
```
cutwater, line 304: Why this function returns empty list only?

cutwater (changes requested):

ansible_events_ui/db/utils/lostream.py
ddonahue007: I think that code legibility is more important than the cut&paste.

tests/integration/api/test_large_object.py
cutwater: In this case maybe it's worth using `random` and `string` modules to generate a random text?

```
import random
import string
...

chars = string.ascii_letters + string.punctuation + string.whitespace
''.join(random.choices(chars, k=1024))
```

ansible_events_ui/db/utils/lostream.py
abaiken: In this situation, I feel that leaving the if else in a one liner is fine. It's still easy to read and in my experience - the test coverage tools will still require you to cover both cases whether or not we break it out into a full blown if/else statement. Conditional expressions are a part of Python general syntax and it is definitely more readable than C-style ternaries, so exploding it into an if-statement is not really necessary IMO.

ansible_events_ui/db/utils/lostream.py
cutwater: @eclarizio here are some examples of existing code:

1. https://github.com/benthomasson/ansible-events-ui/blob/3e47a3387c0404a965c6465aeafb5de286d12f06/ansible_events_ui/db/migrations/versions/202208221331_4c2b57da4a58_project_fkeys.py
2. https://github.com/benthomasson/ansible-events-ui/blob/3e47a3387c0404a965c6465aeafb5de286d12f06/ansible_events_ui/db/migrations/versions/202208221331_4c2b57da4a58_project_fkeys.py
3. https://github.com/benthomasson/ansible-events-ui/blob/3e47a3387c0404a965c6465aeafb5de286d12f06/ansible_events_ui/db/utils/migrations.py

ansible_events_ui/db/utils/lostream.py
cutwater: @abaiken I can accept your opinion on readability, I'll only add that if-expressions are better suited in cases where expression is a requirement (lambda functions for example). Regarding coverage however, it's not that simple. This is an example of how python `coverage` package calculate the code coverage:

https://gist.github.com/cutwater/1bba23b87a96a50304a46b716c568b3f

ansible_events_ui/ruleset.py
```diff
@@ -163,43 +169,43 @@ async def inactivate_rulesets(activation_id):
...
+
+    async with large_object_factory(
+        oid=activation_instance_log_id, session=db, mode="wb"
+    ) as lobject:
+        for buff in iter(lambda: proc.stdout.read(read_chunk_size), b""):
+            await lobject.write(buff)
+            await lobject.flush()  # Does a commit
```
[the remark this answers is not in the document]
Red-HAP: I was conflicted about that.

ansible_events_ui/db/utils/lostream.py
abaiken: Thanks for the example @cutwater. It helped me understand your point about coverage.

tests/integration/api/test_large_object.py
Red-HAP: All that will do is create a same-sized list, just with random letters. Not much different than my constant.
If I wanted to generate random content, I'd use `faker`.

cutwater: I'll clean up the code a bit, won't it?

ansible_events_ui/db/utils/lostream.py
Red-HAP: @eclarizio 
I accept that I'm breaking with the convention that has been established in other parts of the project, but, in my defense:

1. We do not use raw SQL that often and we are apparently planning to rewrite all of the models and queries (unless I'm misunderstanding the 2.0 query style ticket(s))... so the existing raw SQL could be rewritten to utilize `<model>.query` or `<model>(params).save()` or `<model>.query.filter().update()`. forms and the raw SQL in the examples eliminated altogether.
2. The SQL in the `lostream` module is **required** (at least, for now) because none of the async postgresql drivers support a direct large object interface.
3. Right-justifying the SQL code, in my opinion, becomes difficult to maintain with editors that enforce tab-stops and also constrict how much code can be placed on a single line which can impact how readable a query is in the first place.
4. I don't foresee at this time much more raw SQL being included in the project since SQLAlchemy is being utilized in other modules. Creating models for `pg_largeobject_metatata` and `pg_largeobject` are unnecessary as they will _only_ be utilized by the `lostream` module.

tests/integration/api/test_large_object.py
eclarizio: > I'll clean up the code a bit, won't it?

I think a comment will suffice. This is in a test and the only reason I spotted it in the first place was it looked like maybe a private key or secret that when decoded would mean something that we may have to add to a whitelist. I also wasn't sure if there were certain symbols/combinations of symbols that "needed" to be in here for the test. If there's a comment explaining why it exists, I don't feel like it adds much value to import some modules and generate it randomly.

ansible_events_ui/db/utils/lostream.py
cutwater: @Red-HAP 

1. Allow me to give more details on the topic. We already use SQLAlchemy 2.0 style in the project, but we don't enforce it at the moment (`future=True` in the engine configuration), so the database engine still runs in the compatibility mode which has left since I've migrated the codebase from the legacy `databases` wrapper.
   
    The `<model>.query` (which is actually `session.query`) is deprecated since SQLAlchemy 1.4, which introduced new 2.0-style. You can find more details on migration here: https://docs.sqlalchemy.org/en/14/changelog/migration_20.html
   
    The declarative style, is a different story completely. The intention is to use ORM capabilities to utilize data mapper facility and model relationship configuration, which will simplify querying and working with related tables.

2. In most cases the raw SQL is used to call functions (i.e. `lo_get`, `lo_put`, `lo_truncate`, etc.). Isn't it possible to use SQLAlchemy for generating those queries, like I proposed in my other comment:

   ```
   sa.select(sa.func.lo_get(self.oid, self.pos, self.chunk_size))
   ```

   There is one raw SQL query remains in `_verify_large_object`, and referring to item nr. 4, I don't think the change is actually needed ( no need to introduce a model for it ).

3. Can you please elaborate on this or give an example of such editors?

    Considering the previous point, if we're talking about single raw SQL query, I can suggest moving it into a module level constant, which has indentation 0:

    ```
    VERIFY_LARGE_OBJECT_QUERY = """
    select m.oid,
           sum(length(coalesce(d.data, convert_to('', 'utf-8')))) as _length
      from pg_largeobject_metadata m
      left
      join pg_largeobject d
        on d.loid = m.oid
     where m.oid = :_oid
     group
       by m.oid;
    """
    ```

4. See nr. 2

tests/integration/api/test_large_object.py
```diff
@@ -0,0 +1,182 @@
...
+    _write_buffer_bin = _write_buffer.encode("utf-8")
+    assert isinstance(_write_buffer_bin, bytes)
```
Alex-Izquierdo, line 112: What is the purpose of this assertion?

PR #206, opened 2022-10-12
Add GET endpoint to read activation jobs

Dostonbek1:
Add a new API endpoint to retrieve jobs based on activation id.

Resolves: AAP-6201


src/eda_server/db/models/job.py
```diff
@@ -35,6 +36,26 @@
     sa.Column("uuid", postgresql.UUID),
 )
 
+
+activation_job_instances = sa.Table(
```
cutwater, line 40: Is it simply a many-to-many junction table?

src/eda_server/api/activation.py
```diff
@@ -216,6 +216,21 @@ async def delete_activation(
     await db.commit()
 
 
+@router.get(
+    "/api/activations/{activation_id}/job_instances",
+    response_model=List[schema.ActivationJobInstance],
+    operation_id="list_activation_job_instances",
+)
+async def list_activation_job_instances(
```
cutwater, line 224: What's the point of returning only list of job instance IDs?

src/eda_server/db/models/job.py
cutwater, line 40: Since it's a new table, where it's populated?

cutwater (changes requested):
Please add tests.

src/eda_server/db/models/job.py
```diff
@@ -35,6 +36,26 @@
...
+    metadata,
+    sa.Column(
+        "id",
+        sa.Integer,
+        sa.Identity(always=True),
+        primary_key=True,
+    ),
+    sa.Column(
+        "activation_id",
+        sa.ForeignKey("activation.id", ondelete="CASCADE"),
```
cutwater, line 51: Shouldn't it be `nullable=False`?

src/eda_server/db/models/job.py
Dostonbek1: I think it is one-to-many, activation can have multiple job instances but each job instance only belongs to one activation.

cutwater: So why the extra table is needed then?

src/eda_server/api/activation.py
Dostonbek1: I tried to follow the format that we already have for [Activation Instance Job Instances.](https://github.com/ansible/eda-server/blob/cbd41655e3ffe7da886a256ad8bfc2b1e3ad6504/src/eda_server/api/activation.py#L415) Is it wrong?

src/eda_server/db/models/job.py
Dostonbek1: For Activation Instance, its job instance table is populated on activation instance creation but for Activation, we currently don't have that function to populate its job instance table.

Dostonbek1:
This PR was replaced with https://github.com/ansible/eda-server/pull/237.

Dostonbek1:
This PR was replaced with https://github.com/ansible/eda-server/pull/237.

PR #209, opened 2022-10-13
Add podman specific health check when waiting for postgres container

ddonahue007:
[[AAP-6275](https://issues.redhat.com/browse/AAP-6275)] - eda_dev.sh script healthcheck fails on older versions of Podman

### DIFF
--- a/scripts/eda_dev.sh
+++ b/scripts/eda_dev.sh
@@ -94,8 +94,45 @@ start-events-services() {
     log-debug "docker-compose -p eda-server -f tools/docker/docker-compose.yml up -d postgres"
     docker-compose -p eda-server -f tools/docker/docker-compose.yml up -d postgres
 
-    until [ "$(docker inspect -f '{{.State.Health.Status}}' eda-postgres)" == "healthy" ]; do
+    local heath_check=".State.Health.Status"
+
+    #
+    # The below code is needed do to Podman versions pre V4, us a different string variation of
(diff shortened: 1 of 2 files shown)


abaiken (dismissed):

abaiken (approved):

scripts/eda_dev.sh
```diff
@@ -94,8 +94,45 @@ start-events-services() {
...
+    # Podman versions previous to v4 use a different string variation for health check.
+    #
+    # Reference:
+    #      https://github.com/containers/podman/issues/11645
+    #
+    # Default is to use the Docker key value if both Podman and Docker are installed.
+    #
+    local heath_check=".State.Health.Status"
+
+    if which docker &> /dev/null; then
```
Alex-Izquierdo, line 107: This won't work because the "docker" binary is required even if podman is installed. Podman must be checked first.

Alex-Izquierdo (changes requested):

scripts/eda_dev.sh
ddonahue007: Is this due to our requirement of Docker-compose?

ddonahue007: now check to see if podman is installed first.

Alex-Izquierdo (approved):
LGTM

ttuffin (approved):

PR #210, opened 2022-10-13
Adding INSTALL.MD

ttuffin:
Docs for eda-server installation and update dev environment setup instructions.
AAP-6235
AAP-6135

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,11 @@
 # Event Driven Ansible Server
 
+## Installation
+For instructions on how to quickly spin up a preview instance of *eda-server*, please see [INSTALL.md](./INSTALL.md).
+
 ## Setting up a development environment
+
+The below instructions detail how to setup a development environment for _eda-server_.
 
 ### Requirements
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,34 +1,32 @@
 # Event Driven Ansible Server
 
+## Installation
+For instructions on how to quickly spin up a preview instance of *eda-server*, please see [INSTALL.md](./INSTALL.md).
+
 ## Setting up a development environment
 
+The below instructions detail how to setup a development environment for _eda-server_. These instructions should be performed on your local system and not on a remote system.
```
Alex-Izquierdo, line 8: Why shouldn't you be able to use a remote machine as a development environment? I think this sentence is not necessary.

README.md
```diff
@@ -1,34 +1,32 @@
...
-- Python >= 3.8
-- Node >= v16 (**NOTE:** node v17 does not seem to work with websockets)
+- Python >= 3.9
+- python3-pip
+- python3-devel
+- gcc
+- npm >= v16 (**NOTE:** node v17 does not seem to work with websockets)
 
-NOTE podman users (only for mac and linux):
+**NOTE podman users (only for MacOS and Linux):**
```
Alex-Izquierdo, line 22: This section should include as well the tips for the DOCKER_HOST and the service to run the socket.

README.md
ttuffin: The task that builds and starts the UI runs `npm run start:dev`, which opens up a local browser window to the eda page and is only accessible from localhost. @ddonahue007 may be able to elaborate on this further.

Alex-Izquierdo: That's doesn't prevent you from using a remote system.

ddonahue007: When doing hot reload for UI dev work you needs to be running locally.  There are ways to get around this but to be honest I am not sure why you would want to.

ttuffin: Would it be better to remove this sentence and add a note to step 5 to say this command must be run within `graphical.target`?

ddonahue007: I think that we just remove the sentence, If some hits an issue we can address it then I think

ddonahue007 (dismissed):
lgtm

ddonahue007 (approved):

Alex-Izquierdo (approved):
LGTM

PR #223, opened 2022-10-19
Continue refactoring of the project sync service

cutwater:
Changes made:

1. The project files scanning was reworked (preserving the original
   behavior).

   Previous implementation was iterating over project files and
   reading the files multiple times. The directory tree was scanned
   4 times, once per each content type (rulebook, inventory,
   playbook and extra vars).

   New implementation iterates over project files only once. Each
   file is opened and readed only once as well.

2. This allowed to structure the project synchronization better.
   With duplicated code that iterated over the project file being stripped,
   the functions responsible for insertion of files of a specific type
   became pure database query functions, that allows moving them into
   isolated layer in the following update.

3. Use shallow clone (`git clone --depth 1`) of the project repository
   to avoid unnecessary downloading of a project commit history.

4. Archive repository with `git archive` command instead of `tar` to
   store only files tracked by git and not the `.git` repository itself.

### DIFF
--- a/src/eda_server/project.py
+++ b/src/eda_server/project.py
@@ -3,6 +3,9 @@ import logging
 import os
 import shutil
 import tempfile
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, Iterator, Literal, Optional
 
 import sqlalchemy as sa
 import yaml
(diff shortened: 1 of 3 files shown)


src/eda_server/project.py
```diff
@@ -3,114 +3,305 @@
...
+    elif is_inventory_file(content):
+        return ProjectFileKind.INVENTORY
+    elif is_playbook_file(content):
+        return ProjectFileKind.PLAYBOOK
+    else:
+        return ProjectFileKind.EXTRA_VARS
+
+
+def scan_file(path: str) -> Optional[ScanResult]:
+    # TODO(cutwater): Should we support both *.yml and *.yaml extensions ?
```
Alex-Izquierdo, line 206: We should, for sure. IMO. 
.yml is the most used, but .yaml is also widely adopted and is the recommended: https://yaml.org/faq.html

src/eda_server/project.py
```diff
@@ -3,114 +3,305 @@
...
-        logger.exception(filename)
+    for entry in data:
+        if not isinstance(entry, dict):
+            return False
+        if entry.keys() & {"tasks", "roles"}:
+            return True
+    return False
+
+
+def is_inventory_file(data: Any):
```
Alex-Izquierdo, line 269: I know that you are preserving the current behavior but I've found a potential bug, because inventories can be also in other formats: https://docs.ansible.com/ansible/latest/plugins/inventory.html#enabling-inventory-plugins
(in fact, `ini` is widely used). 

Yep, the bug would be more in `scan_file` definition. 
I can open a new ticket if you don't want to do it in this PR.

cutwater: yep, this needs to be addressed. I'd prefer to keep it separately.

cutwater:
@Alex-Izquierdo It was ready for review. I'll rebase it.

PR #224, opened 2022-10-19
[AAP-5061] project pagination

shannon-donahue:
Jira ticket: [AAP-5061](https://issues.redhat.com/browse/AAP-5601)

****Testing Instructions****
1. Bring up Eda-server and add a couple of projects
2. Hit the list projects endpoint: http://localhost:9000/api/projects/
3. Response should be the following format
```
{
    "items": [
        {
            "id": 1,
            "url": "https://github.com/benthomasson/eda-project",
            "name": "testing number 1 today"
        },
        {
            "id": 2,
            "url": "https://github.com/benthomasson/eda-project",
            "name": "testing number 2 today"
        },
    ],
    "total": 2,
    "limit": 50,
    "offset": 0
}
```
4. You can also test varying limit and offset parameters by the following example endpoints
      - http://localhost:9000/api/projects/?limit=2
      -http://localhost:9000/api/projects/?offset=2


src/eda_server/api/project.py
```diff
@@ -30,15 +30,15 @@ async def project_by_name_exists_or_404(db: AsyncSession, project_name: str):
 
 
 @router.get(
-    "/api/projects",
-    response_model=List[schema.ProjectList],
+    "/api/projects/",
```
cutwater, line 33: Please remove trailing slash

src/eda_server/api/project.py
```diff
@@ -30,15 +30,15 @@ async def project_by_name_exists_or_404(db: AsyncSession, project_name: str):
...
+    response_model=LimitOffsetPage[schema.ProjectList],
     operation_id="list_projects",
     tags=["projects"],
 )
 async def list_projects(db: AsyncSession = Depends(get_db_session)):
     query = sa.select(projects.c.id, projects.c.url, projects.c.name)
     result = await db.execute(query)
-    return result.all()
+    return paginate(result.all())
```
cutwater, line 41: This will load all projects from the database and then apply pagination to the result.

src/eda_server/api/project.py
```diff
@@ -30,15 +30,15 @@ async def project_by_name_exists_or_404(db: AsyncSession, project_name: str):
...
 @router.get(
-    "/api/projects",
-    response_model=List[schema.ProjectList],
+    "/api/projects/",
+    response_model=LimitOffsetPage[schema.ProjectList],
     operation_id="list_projects",
     tags=["projects"],
 )
 async def list_projects(db: AsyncSession = Depends(get_db_session)):
     query = sa.select(projects.c.id, projects.c.url, projects.c.name)
```
cutwater, line 39: limit and offset should be passed to the query

cutwater (changes requested):
I don't think `fastapi_pagination` is a good choice. It's probably an easy one.
The design looks heavily inspired by Django DRF and has the following typical problems:

1. It creates strong coupling between layer of database queries and presentation layer.

2. It requires raw query to operate, breaking any presentation, transformation or business logic in the request handler may need to execute after getting query results. Which means it's very limited mechanism.

3. It uses monkey patching on FastAPI routing by adding implicit dependencies.

4. It makes the overall code flow implicit and hard to control or debug.

PR #187, opened 2022-10-04
Store more job details to the job_instance table

hsong-rh:
Store more job details: `playbook/module name, ruleset name, action, hosts and inventory` to the `job_instance` table for the purpose of debug or job rerun.

Resolves: [AAP-5081](https://issues.redhat.com/browse/AAP-5081)
Depends: https://github.com/ansible/ansible-rulebook/pull/166

PR #228, opened 2022-10-24
Add job rerun endpoint

hsong-rh:
Add new endpoint: `post /api/job_instance/{job_instance_id}` to rerun an existing job.

Resolves: [AAP-5082](https://issues.redhat.com/browse/AAP-5082)
Depends: https://github.com/ansible/eda-server/pull/187

### DIFF
--- a/src/eda_server/api/job.py
+++ b/src/eda_server/api/job.py
@@ -123,6 +123,147 @@ async def create_job_instance(
     return {**j.dict(), "id": job_instance_id, "uuid": job_uuid}
 
 
+@router.post(
+    "/api/job_instance/{job_instance_id}",
+    response_model=schema.JobInstanceRead,
+    operation_id="rerun_job_instance",
+)
+async def rerun_job_instance(
+    job_instance_id: int, db: AsyncSession = Depends(get_db_session)
+):
+    query = sa.select(models.job_instances).where(
+        models.job_instances.c.id == job_instance_id
+    )
+    job_instance = (await db.execute(query)).first()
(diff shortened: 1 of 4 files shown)


hsong-rh:
The test failure will be solved when the dependent PR https://github.com/ansible/eda-server/pull/187 is merged

src/eda_server/api/job.py
```diff
@@ -91,6 +91,147 @@ async def create_job_instance(
...
+    query = sa.select(models.extra_vars).where(
+        models.extra_vars.c.id == activation_instance.extra_var_id,
+    )
+    extra_var_row = (await db.execute(query)).first()
+
+    if not extra_var_row:
+        error = f"extra_var {activation_instance.extra_var_id} not found"
+        logger.error(error)
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST, detail=error
```
abaiken, line 163: Should this be a 404 error instead of 400?

hsong-rh: I totally agree the first 400 error: `job_instance not found` should return error code 404. About the others, the missing are due to the broken relations. If we return 404, users may confuse because the job instance do exist.

shannon-donahue (approved):
LGTM

abaiken (approved):
LGTM!

hsong-rh:
Added the permission checking on the rerun endpoint.

shannon-donahue (approved):
tested LGTM!

PR #231, opened 2022-11-01
[UI][AAP-6437] Inventory selection modal

lgalis:
Screenshot from 2022-11-07 10-04-15
Add inventory selection modal


eclarizio (dismissed):
Minor, but I would suggest to squash the commits with the same commit message, so there aren't extras that say the same thing, it's just a bit confusing.

ddonahue007 (approved):
lgtm

shannon-donahue (approved):
LGTM

PR #234, opened 2022-11-02
Shows files on alembic-multiple-heads error

Red-HAP:
This change will automatically show the multiple head files on a multiple heads error from alembic.


Dostonbek1 (approved):
lgtm

PR #239, opened 2022-11-04
Add  env variabels to readme

ddonahue007:
[AAP-6875](https://issues.redhat.com/browse/AAP-6875)

 readme for local dev setup

Need to set the following variable when running in the dev env(using the `dev` tasks)

`export EDA_DEPLOYMENT_TYPE=local`

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,14 +58,20 @@ Install Ansible and `ansible.eda` collection:
 pip install ansible
 ansible-galaxy collection install ansible.eda
 ```
+### 3. Set up env variables
+  Set up the following env variable(s) when testing/running local dev env
 
-### 3. Services
+```shell
+export EDA_DEPLOYMENT_TYPE=local
+```
+
+### 4. Services
 
 **Note:**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,14 +58,20 @@ Install Ansible and `ansible.eda` collection:
 pip install ansible
 ansible-galaxy collection install ansible.eda
 ```
+### 3. Set up env variables
+  Set up the following env variable(s) when testing/running local dev env
 
-### 3. Services
+```shell
+export EDA_DEPLOYMENT_TYPE=local
```
Alex-Izquierdo, line 65: Why? I don't see this envvar used anywhere

ddonahue007: The default is set here: 
https://github.com/ansible/eda-server/blob/main/src/eda_server/config/__init__.py#L23

hsong-rh (approved):
👍

README.md
ddonahue007: I am going tp move this to a note and reference that this is only needed as a temporary setting.

Alex-Izquierdo (approved):

PR #241, opened 2022-11-05
Job events performance

benthomasson:
Improves the performance of processing job events from 200 per second to 4000 per second.

### DIFF
--- a/src/eda_server/api/websocket.py
+++ b/src/eda_server/api/websocket.py
@@ -1,3 +1,4 @@
+import asyncio
 import base64
 import json
 import logging
@@ -230,6 +231,10 @@ async def get_job_instance_id(uuid: str, db: AsyncSession):
     return job_instance.id
 
 
+bulk_job_instance_events = asyncio.Queue()
+bulk_job_instance_hosts = asyncio.Queue()
+
+
 async def handle_ansible_rulebook(data: dict, db: AsyncSession):
     event_data = data.get("event", {})
     if event_data.get("stdout"):
@@ -253,20 +258,19 @@ async def handle_ansible_rulebook(data: dict, db: AsyncSession):
             json.dumps(["Stdout", {"stdout": event_data.get("stdout")}]),
         )
 
-    return
-
     created = event_data.get("created")
     if created:
         created = datetime.strptime(created, "%Y-%m-%dT%H:%M:%S.%f")
 
-    query = insert(models.job_instance_events).values(
-        job_uuid=event_data.get("job_id"),
-        counter=event_data.get("counter"),
-        stdout=event_data.get("stdout"),
-        type=event_data.get("event"),
-        created_at=created,
+    await bulk_job_instance_events.put(
+        dict(
+            job_uuid=event_data.get("job_id"),
+            counter=event_data.get("counter"),
+            stdout=event_data.get("stdout"),
+            type=event_data.get("event"),
+            created_at=created,
+        )
     )
-    await db.execute(query)
 
     event = event_data.get("event")
     if event and event in [item.value for item in host_status_map]:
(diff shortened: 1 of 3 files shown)


benthomasson:
Events per second:  4238.371749744774

src/eda_server/api/websocket.py
```diff
@@ -220,14 +223,64 @@ async def handle_workers(websocket: WebSocket, data: dict, db: AsyncSession):
...
+
+
+async def insert_job_instance_events(
+    bulk_job_instance_events: asyncio.Queue, db_session_factory
+):
+    async with db_session_factory() as db:
+        query = insert(models.job_instance_events).values(
+            [
+                await bulk_job_instance_events.get()
+                for i in range(bulk_job_instance_events.qsize())
```
Red-HAP, line 242: Shouldn't this reference the `queue` attribute?
```suggestion
                await bulk_job_instance_events.queue.get()
                for i in range(bulk_job_instance_events.queue.qsize())
```

src/eda_server/api/websocket.py
```diff
@@ -220,14 +223,64 @@ async def handle_workers(websocket: WebSocket, data: dict, db: AsyncSession):
...
+
+
+async def insert_job_instance_hosts(
+    bulk_job_instance_hosts: asyncio.Queue, db_session_factory
+):
+    async with db_session_factory() as db:
+        query = insert(models.job_instance_hosts).values(
+            [
+                await bulk_job_instance_hosts.get()
+                for i in range(bulk_job_instance_hosts.qsize())
```
Red-HAP, line 256: Shouldn't this also reference the `queue` attribute?
```suggestion
                await bulk_job_instance_hosts.queue.get()
                for i in range(bulk_job_instance_hosts.queue.qsize())
```

Red-HAP (changes requested):

src/eda_server/api/websocket.py
benthomasson: Yes, I'll take a look at it.

src/eda_server/api/websocket.py
```diff
@@ -220,14 +223,64 @@ async def handle_workers(websocket: WebSocket, data: dict, db: AsyncSession):
...
+            ]
+        )
+        await db.execute(query)
+        await db.commit()
+
+
+bulk_job_instance_events = Batcher(insert_job_instance_events)
+bulk_job_instance_hosts = Batcher(insert_job_instance_hosts)
```
cutwater, line 264: Global dependency

src/eda_server/batch.py
```diff
@@ -0,0 +1,34 @@
...
+        self.max_batch_size = max_batch_size
+        self.batch_timeout = batch_timeout
+        self.processor_fn = processor_fn
+
+    def start(self, get_db_session_factory):
+        self.get_db_session_factory = get_db_session_factory
+        _ = asyncio.get_event_loop()
+        asyncio.create_task(self._batcher())
+
+    async def _batcher(self):
```
cutwater, line 20: The logic here looks not very intuitive and probably not performant as well.
It runs a loop to check the queue size with fixed intervals and then passes the queue to the function that consumes messages from it. The queue size is unlimited which may lead to the back pressure problem.

While this somehow works because queue size is known and the queue access doesn't need synchronization, I'd re-write it in a different manner.

1. Use limited queue size to avoid uncontrolled growth of a queue.
2. The `processor_fn` callback should expect full batch: a sequence (e.g. list) of items to be processed.
2. The Batcher consumes the queue with a timeout:

```
# PSEUDOCODE #

while True:
    try:
        item = await asyncio.wait(queue.get(), timeout - elapsed_time)
    except TimeoutError:
        # flush
    batch.append(item)
    if len(batch) > max_batch_size:
        # flush
```

cutwater (changes requested):

PR #242, opened 2022-11-07
Fix the missing user resource type

hsong-rh:
Add the missing `user` ResourceType in the migration file.


shannon-donahue (approved):
LGTM

PR #245, opened 2022-11-08
[UI][AAP-6645] Updated left hand nav with expandable Management section and Audit view placeholder

lgalis:
Updated left hand nav with expandable Management section and Audit view placeholder

Screenshot from 2022-11-08 07-07-21


ddonahue007 (approved):
lgtm

PR #246, opened 2022-11-08
Changes to handle bash version 3.x and above

ddonahue007:
[AAP-6939](https://issues.redhat.com/browse/AAP-6939)

We now require bash version 3.x and above, instead of bash 5.x and above


abaiken (approved):

Dostonbek1 (approved):
LGTM

PR #255, opened 2022-11-11
Add error handling for activation creation

Dostonbek1:
Add error handling for activation creation when related objects don't exist.

Testing:
1. Send POST request to both `http://localhost:9000/api/activations` and `http://localhost:9000/api/activation_instance` with the following request body, where an ID is passed for `rulebook_id` which doesn't exist:
```
{
  "name": "test",
  "rulebook_id": 100,
  "inventory_id": 1,
  "extra_var_id": 1,
  "working_directory": "/tmp"
}
```
2. Verify the request above returns a response similar to the following:
```
{
    "detail": "Rulebook with ID=100 does not exist."
}
```
3. You can try the same process for `inventory` and `extra_var` where the ID's don't exist, which should return similar error message. 

Resolves: AAP-6098

### DIFF
--- a/src/eda_server/api/activation.py
+++ b/src/eda_server/api/activation.py
@@ -65,10 +65,10 @@ async def create_activation(
     try:
         result = await db.execute(query)
     except sa.exc.IntegrityError as e:
-        errorInfo = e.orig.args[0].split("\n")
+        error_info = e.orig.args[0].split("\n")
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            detail=str(errorInfo[1]),
+            detail=str(error_info[1]),
         )
     await db.commit()
     (id_,) = result.inserted_primary_key
@@ -283,10 +283,10 @@ async def create_activation_instance(
     try:
         result = await db.execute(query)
     except sa.exc.IntegrityError as e:
-        errorInfo = e.orig.args[0].split("\n")
+        error_info = e.orig.args[0].split("\n")
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            detail=str(errorInfo[1]),
+            detail=str(error_info[1]),
         )
     await db.commit()
     id_, large_data_id = result.first()
--- a/src/eda_server/api/activation.py
+++ b/src/eda_server/api/activation.py
@@ -59,7 +59,6 @@ async def dependent_object_exists_or_exception(db: AsyncSession, activation):
     for dep_object, object_id in zip(
         activation_dependent_objects, dep_object_ids
     ):
-        await db.rollback()
         object_exists = await db.scalar(
             sa.select(sa.exists().where(dep_object[0].c.id == object_id))
         )
@@ -97,6 +96,7 @@ async def create_activation(
     try:
         result = await db.execute(query)
     except sa.exc.IntegrityError:
+        await db.rollback()
         await dependent_object_exists_or_exception(db, activation)
 
     await db.commit()
@@ -311,6 +311,7 @@ async def create_activation_instance(
     try:
         result = await db.execute(query)
     except sa.exc.IntegrityError:
+        await db.rollback()
         await dependent_object_exists_or_exception(db, a)
     await db.commit()
     id_, large_data_id = result.first()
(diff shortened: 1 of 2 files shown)


abaiken (dismissed):
Tested 👍

src/eda_server/api/activation.py
```diff
@@ -26,9 +27,30 @@
 router = APIRouter(tags=["activations"])
 
 
+def activation_unprocessable_entity_exception(e):
+    error_detail = e.orig.args[0].split("\n")[1]
+    object_id = re.findall(r"\d+", error_detail)[0]
+    dependent_objects = ["rulebook", "inventory", "extra_var", "project"]
+    for object in dependent_objects:
+        if object in error_detail:
+            return JSONResponse(
```
Alex-Izquierdo, line 36: Should it return a generic exception if dependent object is not found?

src/eda_server/api/activation.py
```diff
@@ -26,9 +27,30 @@
 router = APIRouter(tags=["activations"])
 
 
+def activation_unprocessable_entity_exception(e):
+    error_detail = e.orig.args[0].split("\n")[1]
+    object_id = re.findall(r"\d+", error_detail)[0]
```
Alex-Izquierdo, line 32: should it be inside a try block? otherwise it could return an IndexError

src/eda_server/api/activation.py
Dostonbek1: We are using this approach when we hit 500 docker error on activation creation instance, so I was trying to keep it consistent with that one.

cutwater (changes requested):
You shouldn't parse the exception text to determine the exception reason.
Either perform a select query before to determine if the dependent objects exist or in the event of exception

src/eda_server/api/activation.py
```diff
@@ -35,6 +68,8 @@ async def create_activation(
     activation: schema.ActivationCreate,
     db: AsyncSession = Depends(get_db_session),
 ):
+    await dependent_object_exists_or_exception(db, activation)
```
cutwater, line 71: Since this function performs 4 SQL queries, it may be reasonable to call it in response of `IntegrityError` exception to clarify the exception reason.

src/eda_server/api/activation.py
```diff
@@ -26,6 +26,39 @@
...
+        activation.extra_var_id,
+        activation.project_id,
+    ]
+
+    for dep_object, object_id in zip(
+        activation_dependent_objects, dep_object_ids
+    ):
+        dependent_object = (
+            await db.execute(
+                sa.select(dep_object[0]).where(dep_object[0].c.id == object_id)
```
cutwater, line 49: No need to select entire record to check its existence. Use `SELECT EXISTS` instead.

cutwater (changes requested):

src/eda_server/api/activation.py
```diff
@@ -41,6 +41,38 @@
...
+        activation.rulebook_id,
+        activation.inventory_id,
+        activation.extra_var_id,
+        activation.project_id,
+    ]
+
+    for dep_object, object_id in zip(
+        activation_dependent_objects, dep_object_ids
+    ):
+        await db.rollback()
```
cutwater, line 62: No need to call rollback in a loop.

src/eda_server/api/activation.py
```diff
@@ -65,10 +97,8 @@ async def create_activation(
     try:
         result = await db.execute(query)
     except sa.exc.IntegrityError:
-        raise HTTPException(
-            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            detail="Unprocessable Entity.",
-        )
+        await dependent_object_exists_or_exception(db, activation)
```
cutwater, line 100: You can call rollback one here, before calling this function.

src/eda_server/api/activation.py
```diff
@@ -279,7 +308,10 @@ async def create_activation_instance(
             models.activation_instances.c.large_data_id,
         )
     )
-    result = await db.execute(query)
+    try:
+        result = await db.execute(query)
+    except sa.exc.IntegrityError:
+        await dependent_object_exists_or_exception(db, a)
```
cutwater, line 314: And here.

cutwater (changes requested):

cutwater (approved):

cutwater (approved):

abaiken (approved):

PR #259, opened 2022-11-14
Docker fixes

benthomasson:


requirements.txt
```diff
@@ -1,3 +1,4 @@
+wheel
```
cutwater, line 1: Why the `wheel` package is added to the dependencies list ?

PR #261, opened 2022-11-15
[AAP-5991] Add API permission checks

cutwater:
* Modify test suite to make permission checks testing easier.
* Add permission checks and update tests for existing API endpoints.
* Configure `app` fixture for API tests to authenticate as admin user.
* Move `test_create_rulebook` test case to respective tests file.
* Move `admin_user` fixture to API's `conftest.py`.
* Add `check_permission_spy` fixture to test permission checks.
* Add new resource types to sample `initial_data.yml`

### DIFF
--- a/src/eda_server/api/activation.py
+++ b/src/eda_server/api/activation.py
@@ -26,6 +26,7 @@ from fastapi.responses import JSONResponse
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from eda_server import schema
+from eda_server.auth import requires_permission
 from eda_server.config import Settings, get_settings
 from eda_server.db import models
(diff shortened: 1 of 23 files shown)


hsong-rh:
@cutwater I tested, now APIs works without that transaction begun error. But when I logged in with `bob@example.com`, I got 403 forbidden error when reading projects, actions, etc. Bob supposes to take `default` role and have permissions to read them, right?

cutwater:
@hsong-rh The default role as of now is configured via environment variables. It's disabled by default and you need to enable it. I'm working on a separate patch that will change this behavior.

hsong-rh (approved):
LGTM

PR #274, opened 2022-11-22
[AAP-5245] Improve test coverage for APIs permission checking

hsong-rh:
Add more test cases for permission checks on the endpoints of `Action Instance`, `Extra Var`, `Playbook`, `Role`, `Job`, `Task` and `Rulebook`.

Resolves: [AAP-5245](https://issues.redhat.com/browse/AAP-5245)

### DIFF
--- a/tests/integration/api/test_activation.py
+++ b/tests/integration/api/test_activation.py
@@ -40,6 +40,16 @@ TEST_ACTIVATION = {
     "project_id": 1,
 }
 
+TEST_ACTIVATION_INSTANCE = {
+    "name": "test-activation_instance",
+    "rulebook_id": 1,
+    "inventory_id": 1,
+    "extra_var_id": 1,
+    "working_directory": "/tmp",
+    "execution_environment": "quay.io/aizquier/ansible-rulebook",
+    "project_id": 1,
+}
+
 TEST_EXTRA_VAR = """
 ---
 collections:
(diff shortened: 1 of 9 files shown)


tests/integration/api/test_activation.py
```diff
@@ -40,6 +40,16 @@
     "project_id": 1,
 }
 
+TEST_ACTIVATION_INSTANCE = {
+    "name": "test-activation_instance",
+    "rulebook_id": 1,
+    "inventory_id": 1,
+    "extra_var_id": 1,
+    "working_directory": "/tmp",
+    "execution_environment": "quay.io/aizquier/ansible-rulebook",
```
shannon-donahue, line 49: Just a quick question on my end. Is there any sort of generic execution environment variable we could set up for testing, or is this the standard environment we have decided as a testing execution environment?

hsong-rh: @shannon-donahue No, there is no such a standard execution environment. For testing, it can be any variable name.

Alex-Izquierdo: This is the execution environment by default, which is an image automatically built from the official repo. 
https://github.com/ansible/eda-server/blob/main/src/eda_server/schema/activation.py#L36

cutwater (commented):
The description states that this PR adds missing permission checks, but from the PR diff itself I can't see that any of the permission checks were updated.

cutwater (commented):
Optional: It would be nice to start moving queries into `eda_server.db.sql` package.

hsong-rh:
> The description states that this PR adds missing permission checks, but from the PR diff itself I can't see that any of the permission checks were updated.

Description is updated to clarify

hsong-rh:
> Optional: It would be nice to start moving queries into `eda_server.db.sql` package.

Need many changes, better to open another ticket/PR to do it.

tests/integration/api/test_activation.py
shannon-donahue: Thank you both for the clarification

shannon-donahue:
Gave this another test this morning. I ran pytest locally and I am hitting the following error in the test_activation.py phase of testing. 
```
File "/Ansible/eda-server/src/eda_server/api/websocket.py", line 165, in handle_workers
models.projects.c.id == activation.project_id
AttributeError: 'NoneType' object has no attribute 'project_id'
```

hsong-rh:
> Optional: It would be nice to start moving queries into `eda_server.db.sql` package.



> Gave this another test this morning. I ran pytest locally and I am hitting the following error in the test_activation.py phase of testing.
> 
> ```
> File "/Ansible/eda-server/src/eda_server/api/websocket.py", line 165, in handle_workers
> models.projects.c.id == activation.project_id
> AttributeError: 'NoneType' object has no attribute 'project_id'
> ```

@shannon-donahue I rebased and unable to reproduce the above error. Which test case failed in the `test_activation.py` or all test cases?

shannon-donahue:
> 
It may be something on my local machine or due to the fact I am running a clean empty DB. investigating a little further. The logs did not print out which test but judging from how many ran its on a create activation instance test. The error is on the websocket worker handler for a none existent project ID. 

Just to have the steps laid out I:
1. Cleaned the services - `task dev:services:clean`
2. brought up development - `task dev:all:start`
3. ran `pytest`

PR #282, opened 2022-12-01
Update extravar and playbook urls

lgalis:
Fixed the playbook and extravars get urls

 - Fixes the list of playbook and extravars in the New Activation and New Job forms following the RBAC PR

### DIFF
--- a/ui/src/app/API/Extravar.ts
+++ b/ui/src/app/API/Extravar.ts
@@ -1,7 +1,7 @@
 import { getAxiosInstance } from '@app/API/baseApi';
 import { AxiosResponse } from 'axios';
 
-const extravarsEndpoint = '/api/extra_vars';
+const extravarsEndpoint = '/api/extra_vars/';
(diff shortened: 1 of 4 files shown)


ttuffin:
We actually have a bug open (AAP-7409) to remove the trailing slashes on the API side in order to align with the [API design guide](https://github.com/ansible/eda-server/blob/main/README-API.md). So I think this change should not go ahead?

Alex-Izquierdo:
Indeed, if this bug is accepted: https://issues.redhat.com/browse/AAP-7409 (I see no reason not to.) This PR would be unnecessary.

lgalis:
@ttuffin, @Alex-Izquierdo - AAP-7409 was not yet assigned - I have asked in the team chat to see if someone can work on this. If not - we can merge this PR and revert when the API is fixed

lgalis:
Closing as it was fixed on the api side https://github.com/ansible/eda-server/pull/284

PR #287, opened 2022-12-06
Uses ruleset.sources for port, other config

Red-HAP:
## Jira Ticket

[AAP-5433](https://issues.redhat.com/browse/AAP-5433)

### Docker Deployments Only

If sources are available in the ruleset, use the port information in the source for websocket port allocation. 

If sources are not available, fallback to original creation method

Uses the `sources` column from the `ruleset` table.

Queries have been moved to their own file. An insert followed by a select has been combined into one query using `cte()`

## Testing

Create a new rulebook that has ruleset sources. The port from the ruleset source should allocated for the docker container. The logs will show the host and port used.

Create a new rulebook that does not have ruleset sources. The port from the ruleset source should allocated for the docker container. The logs will show the host and port used. These should match those in the settings.


src/eda_server/config/enums.py
```diff
@@ -0,0 +1,25 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""EDA Server enum definitions."""
+
+from enum import Enum
+
+
+class DeploymentTypeEnum(Enum):
```
cutwater, line 20: Minor: I think using `Enum` suffix for enums is redundant.

src/eda_server/config/enums.py
```diff
@@ -0,0 +1,25 @@
...
+    LOCAL = "local"
+    DOCKER = "docker"
+    PODMAN = "podman"
+    K8S = "k8s"
+    KUBERNETES = "kubernetes"
```
cutwater, line 25: What is the difference between `k8s` and `kubernetes` ?

tests/integration/api/test_ruleset.py
```diff
@@ -0,0 +1,129 @@
...
+rulesets = ""
+ruleset_sources = []
+inventory = ""
+extravars = ""
+working_directory = "/tmp"
+host = "test-eda-server"
+port = 15000
+
+
+class TstDocker(aiodocker.docker.Docker):
```
cutwater, line 21: Minor: Naming. Please try to avoid unnecessary name shortenings.

tests/integration/api/test_ruleset.py
```diff
@@ -0,0 +1,129 @@
...
+    def __init__(self, *args, **kwargs):
```
cutwater, line 22: If this is a mock class, why does it inherit real `aiodocker.docker.Docker` class?

tests/integration/api/test_ruleset.py
```diff
@@ -0,0 +1,129 @@
...
+class TstProc(asyncio.subprocess.Process):
+    def __init__(self, *args, **kwargs):
+        self.pid = 19999
+
+
+def test_resolve_activation_function():
+    expected = "create_websocket_docker_activation"
+    src = {}
+    f = ruleset.resolve_activation_function(src)
+    assert f.__name__ == expected
```
cutwater, line 43: You can check identity against the function itself, not it's name

```
f is create_websocket_docker_activation
```

src/eda_server/ruleset.py
```diff
@@ -73,13 +78,194 @@ def ensure_directory(directory):
...
+        raise
+
+    return (docker, container)
+
+
+def resolve_activation_function(source: Dict) -> Callable:
+    # TODO: React to source type
+    activation_type = "websocket"
+    activation_func_name = f"create_{activation_type}_docker_activation"
+    activation_func = globals()[activation_func_name]
```
cutwater, line 208: Please don't do that. If you need to resolve a function by it's name, create a dictionary of names and references to functions.

cutwater (changes requested):

src/eda_server/config/enums.py
Red-HAP: JIC someone wants to spell it out. K8S was the original, IIRC.

tests/integration/api/test_ruleset.py
Red-HAP: Was probably going to check something with `isinstance` that didn't become realized.

tests/integration/api/test_ruleset.py
Red-HAP: Yeah. I can fix that.

ddonahue007 (approved):
Testing this in the docker configuration seems to work as describe

